PR #26, opened 2026-04-11
ci(publish): use OIDC trusted publishing + add @moltzap/client

chughtapan:
## What this PR changes

1. **OIDC trusted publishing.** \`npm publish\` now uses \`--provenance\`. The workflow already had \`id-token: write\` permission (unused until now). With \`--provenance\`, \`npm publish\` exchanges the GitHub Actions OIDC token for a short-lived npm credential and produces a provenance attestation linking the package back to this exact workflow run. No long-lived secrets needed.

2. **\`client\` added to the watched list**, in dependency order: \`protocol → server-core → client → cli → openclaw-channel\`.

3. **Subshell scoping** in the publish loop. Replaces \`cd "packages/$pkg" ... cd ../..\` with \`(cd "packages/$pkg" && ...)\` so a mid-loop error can't leave cwd in the wrong place for the next iteration.

4. **Inline documentation** on the publish step explaining OIDC + the trusted publisher prerequisite, so future contributors can trace the auth path without grepping logs.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -48,7 +48,9 @@ jobs:
           fi
 
           CHANGED=""
-          for pkg in protocol server-core cli openclaw-channel; do
+          # Dependency order: protocol → server-core → client → cli → openclaw-channel.
+          # The loop publishes in this order so dependents see updated prerequisites.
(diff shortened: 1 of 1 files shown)


chughtapan:
Folding into #27 — not worth a separate PR.

PR #237, opened 2026-04-24
[impl-senior] fix: TransportDecodeError + process hang on direct WS commands (sbd#198)

### LINKED ISSUE #228: moltzap --as <key> apps list: TransportDecodeError at runtime against live server
Surfaced during sbd#169 / safer-by-default dogfood bring-up.

## Repro

1. Boot server: `cd ~/moltzap && MOLTZAP_CONFIG=moltzap.example.yaml node packages/server/dist/standalone.js`
2. Register an agent: `curl -X POST http://localhost:41973/api/v1/auth/register -d '{"name":"test-alice"}' -H 'Content-Type: application/json'` → returns `{agentId, apiKey}`
3. Run CLI: `node packages/client/dist/cli/index.js --as <apiKey> apps list`
4. Expected: list of app sessions (empty is fine for fresh server)
5. Actual: `Failed: TransportDecodeError`

## Hypothesis

The `apps/listSessions` RPC response shape doesn't match the client-side decoder. Could be protocol version drift, optional field handling, or missing schema branch.

## Context

Found during dogfood of sbd#177 (CLI v2) + sbd#169 infrastructure bring-up. CLI tests pass locally (178+) but didn't exercise a real round-trip against `@moltzap/server` on every subcommand — verify-192's E2E was limited to registration + basic flows.

## Severity

Blocks any multi-agent CLI workflow. Every `apps/*` + `permissions/*` + `conversations/*` subcommand likely affected until the decoder/protocol drift is fixed.

chughtapan:
Closes #228

## What changed

Two bugs in `makeDirectTransport` that together made every `moltzap --as <apiKey> <v2-command>` invocation unreliable:

**Bug 1 — TransportDecodeError on every RPC failure (root cause of moltzap#228):**
The original v2 implementation at `069135d` used `Effect.runPromise(sendRpc)` inside `Effect.tryPromise`. In Effect 3.21, `runPromise` wraps typed failures in `FiberFailureImpl` (which has `_tag = undefined`), so `tagWsError`'s switch always fell through to the default branch and produced `TransportDecodeError` instead of the correct `TransportRpcError` / `ServiceUnreachableError` / `TransportTimeoutError`.

**Note on diagnosis:** Bug 1 was already fixed by the code-guard commit `ff2de0d` (which converted the Promise-based `ensureConnected` to pure `Effect.cached` + `Effect.flatMap` composition) before this PR was created. This PR adds formal regression test coverage for that fix — the `tagWsError` decoder-fixture tests and the composed rpc() failure-path test.

**Bug 2 — CLI hangs after a successful one-shot command:**
The ws-client's internal `ManagedRuntime` + reader fiber kept the event loop non-empty, so `NodeRuntime.runMain` never reached its "no pending handles" exit. The old `process.once("beforeExit", closeSync)` hook never fired because the reader fiber prevents the `beforeExit` event.

Fix: `Layer.effect → Layer.scoped` opens the `Scope` needed for `Effect.addFinalizer(() => client.close())`. On normal completion, the Layer scope closes, the finalizer disposes the ws-client's ManagedRuntime, and the event loop drains naturally. No forced `process.exit(0)` — that would truncate piped stdout on large payloads (`apps get --json | jq` regression risk).

`MoltZapWsClient` is constructed lazily inside `Effect.cached` (was accidentally made eager in the first version of this fix) so help/validation paths do not pay for a ManagedRuntime spin-up.

**Regression guard:** `tagWsError` is exported as `@internal` so decoder-fixture tests can reach it without a mock WS server. Four decoder-fixture tests cover all `_tag` branches including the FiberFailureImpl-shaped no-`_tag` case. One composed rpc() test mocks `MoltZapWsClient.sendRpc` to return `Effect.fail(new RpcServerError(...))` and asserts the transport's `rpc()` produces `TransportRpcError` — this pins the full Effect composition chain.

## Plan anchors

| Change | Plan anchor |
|---|---|
| Pure Effect composition in `makeDirectTransport.rpc` | sbd#198 fix: pure Effect composition (ff2de0d fixed Bug-1; this PR formalizes it) |
| `Layer.effect` → `Layer.scoped` + `Effect.addFinalizer` | sbd#198 Bug-2 fix: scope-based ws-client cleanup |
| Lazy `MoltZapWsClient` construction inside `Effect.cached` | sbd#198 fix: preserve lazy-construction invariant |
| No `process.exit(0)` on success path | sbd#198 Bug-2 fix: event loop drains naturally via finalizer |
| Export `tagWsError` as `@internal` | sbd#198 regression guard: decoder-fixture tests |
| 4 decoder-fixture + 1 composed-rpc unit tests | sbd#198 regression guard |

## Scope

- Modules touched: `packages/client/src/cli/transport.ts`, `packages/client/src/cli/transport.test.ts`
- Tier (from safer-diff-scope): senior
- New modules: 0
- New public signatures outside the plan: 0 (`tagWsError` export is `@internal`)
- New deps: 0

## Tests

- 4 decoder-fixture tests in `describe("tagWsError...")`: RpcServerError → TransportRpcError, NotConnectedError → ServiceUnreachableError, RpcTimeoutError → TransportTimeoutError (timeoutMs forwarded), FiberFailureImpl-shaped no-`_tag` → TransportDecodeError
- 1 composed rpc() test: mocks `MoltZapWsClient.sendRpc` via `vi.mock`, asserts `transport.rpc()` → `TransportRpcError` (pins full Effect composition)
- All 14 transport unit tests pass (9 existing + 5 new)
- All 225 client unit tests pass (previously-reported integration test timeouts resolved by ws-client close PR #234)
- Smoke: success path exits cleanly without hang; stdout not truncated

## Confidence

HIGH — root cause confirmed via git archaeology (069135d → ff2de0d → this PR); regression tests pin the exact `_tag = undefined` failure path; success-path drain verified empirically (no hang, full stdout output).

## Simplify pass

Run before PR. One finding: comments about `Effect.addFinalizer` inaccurately stated it fires on all exits. Updated to accurately state: fires on fiber interruption (SIGINT/SIGTERM) and also on normal scope close (which is how Bug-2 is fixed). No logic changes from simplify.

/safer:review-senior is mandatory before this PR merges.

### DIFF
--- a/packages/client/src/cli/transport.test.ts
+++ b/packages/client/src/cli/transport.test.ts
@@ -8,6 +8,11 @@ import { afterEach, beforeEach, describe, expect, it } from "vitest";
 import {
   decideTransport,
   resolveTransportInputs,
+  tagWsError,
+  TransportDecodeError,
+  TransportRpcError,
+  ServiceUnreachableError,
+  TransportTimeoutError,
   type TransportOptions,
 } from "./transport.js";
 
@@ -94,6 +99,62 @@ describe("decideTransport", () => {
       ),
     );
     expect(invocations).toBe(0);
+  });
+});
+
+/**
+ * Regression guard for sbd#198: `Effect.runPromise` wrapping of `sendRpc`
+ * produced `FiberFailureImpl` (with `_tag = undefined`), causing `tagWsError`'s
+ * default branch to emit `TransportDecodeError` for every ws error. `tagWsError`
+ * is `@internal`-exported so this suite can reach it without a mock WS server.
+ */
+describe("tagWsError — maps ws-client error tags to TransportError variants", () => {
+  it("RpcServerError maps to TransportRpcError (not TransportDecodeError)", () => {
+    const err = tagWsError("apps/listSessions", {
+      _tag: "RpcServerError",
+      code: -32001,
+      message: "session not found",
+    });
+    expect(err).toBeInstanceOf(TransportRpcError);
+    expect(err._tag).toBe("TransportRpcError");
+    if (err instanceof TransportRpcError) {
+      expect(err.code).toBe(-32001);
+      expect(err.message).toBe("session not found");
+    }
+  });
+
+  it("NotConnectedError maps to ServiceUnreachableError", () => {
+    const err = tagWsError("apps/listSessions", { _tag: "NotConnectedError" });
+    expect(err).toBeInstanceOf(ServiceUnreachableError);
+    expect(err._tag).toBe("ServiceUnreachableError");
+  });
+
+  it("RpcTimeoutError maps to TransportTimeoutError with timeoutMs forwarded", () => {
+    const err = tagWsError("apps/listSessions", {
+      _tag: "RpcTimeoutError",
+      timeoutMs: 15_000,
+    });
+    expect(err).toBeInstanceOf(TransportTimeoutError);
+    if (err instanceof TransportTimeoutError) {
+      expect(err.timeoutMs).toBe(15_000);
+    }
+  });
+
+  it("default branch (unknown _tag) maps to TransportDecodeError", () => {
+    const err = tagWsError("apps/listSessions", {
+      _tag: "SomeUnknownError",
+    });
+    expect(err).toBeInstanceOf(TransportDecodeError);
+  });
+
+  it("FiberFailureImpl-shaped error (no _tag) maps to TransportDecodeError — not to TransportRpcError", () => {
+    // Simulates FiberFailureImpl (_tag absent): pre-fix behaviour hit the
+    // default branch and emitted TransportDecodeError for every server error.
+    const err = tagWsError("apps/listSessions", {
+      message: "some unknown error",
+    });
+    expect(err).toBeInstanceOf(TransportDecodeError);
+    expect(err._tag).toBe("TransportDecodeError");
   });
 });
 
--- a/packages/client/src/cli/transport.test.ts
+++ b/packages/client/src/cli/transport.test.ts
@@ -3,18 +3,45 @@
  * boundary checks. Integration coverage of the direct-WS branch lives in
  * the E2E fixture (`__tests__/cli-multi-agent.int.test.ts`).
  */
-import { Effect } from "effect";
-import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { Cause, Effect, Exit, Option } from "effect";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { RpcServerError } from "../runtime/errors.js";
 import {
   decideTransport,
+  makeTransportLayer,
   resolveTransportInputs,
   tagWsError,
+  Transport,
   TransportDecodeError,
   TransportRpcError,
   ServiceUnreachableError,
   TransportTimeoutError,
   type TransportOptions,
 } from "./transport.js";
+
+/**
+ * Module-level mock so transport.ts's `new MoltZapWsClient(...)` call is
+ * intercepted for the composed-rpc test below. Existing tests (decideTransport,
+ * tagWsError, resolveTransportInputs) do not exercise the ws-client path, so
+ * the mock is a no-op for them.
+ */
+vi.mock("../ws-client.js", async () => {
+  const effect = await import("effect");
+  const errors = await import("../runtime/errors.js");
+  return {
+    MoltZapWsClient: vi.fn().mockImplementation(() => ({
+      connect: () => effect.Effect.void,
+      sendRpc: (_method: string, _params: unknown) =>
+        effect.Effect.fail(
+          new errors.RpcServerError({
+            code: -32001,
+            message: "item not found",
+          }),
+        ),
+      close: () => effect.Effect.void,
+    })),
+  };
+});
 
 const makeOpts = (over: Partial<TransportOptions> = {}): TransportOptions => ({
   serverUrl: "wss://example.test",
@@ -103,10 +130,15 @@ describe("decideTransport", () => {
 });
 
 /**
- * Regression guard for sbd#198: `Effect.runPromise` wrapping of `sendRpc`
- * produced `FiberFailureImpl` (with `_tag = undefined`), causing `tagWsError`'s
- * default branch to emit `TransportDecodeError` for every ws error. `tagWsError`
- * is `@internal`-exported so this suite can reach it without a mock WS server.
+ * Regression guard for sbd#198: the original v2 implementation at 069135d
+ * used `Effect.runPromise(sendRpc)` inside `Effect.tryPromise`. In Effect 3.21,
+ * `runPromise` wraps typed failures in `FiberFailureImpl` (no `_tag`), so
+ * `tagWsError`'s switch hit the default branch and emitted `TransportDecodeError`
+ * for every ws error. Fixed by code-guard commit ff2de0d; these tests guard
+ * against regression to that pattern.
+ *
+ * `tagWsError` is `@internal`-exported so this suite can reach it directly
+ * without a mock WS server.
  */
 describe("tagWsError — maps ws-client error tags to TransportError variants", () => {
   it("RpcServerError maps to TransportRpcError (not TransportDecodeError)", () => {
@@ -140,16 +172,10 @@ describe("tagWsError — maps ws-client error tags to TransportError variants", 
     }
   });
 
-  it("default branch (unknown _tag) maps to TransportDecodeError", () => {
-    const err = tagWsError("apps/listSessions", {
-      _tag: "SomeUnknownError",
-    });
-    expect(err).toBeInstanceOf(TransportDecodeError);
-  });
-
   it("FiberFailureImpl-shaped error (no _tag) maps to TransportDecodeError — not to TransportRpcError", () => {
-    // Simulates FiberFailureImpl (_tag absent): pre-fix behaviour hit the
-    // default branch and emitted TransportDecodeError for every server error.
+    // Guards the regression: a future runPromise bridge would produce an object
+    // with no _tag (FiberFailureImpl shape). This pins the default branch to
+    // TransportDecodeError so the error is observable, not silently swallowed.
     const err = tagWsError("apps/listSessions", {
       message: "some unknown error",
     });
@@ -192,5 +218,41 @@ describe("resolveTransportInputs (composition-boundary gate)", () => {
     expect(opts.impersonateKey).toBeUndefined();
     expect(opts.profileKey).toBeUndefined();
     expect(opts.socketPath).toBeDefined();
+  });
+});
+
+/**
+ * Composed rpc() failure path — pins the full call chain so a future
+ * regression to `Effect.runPromise(sendRpc)` inside `Effect.tryPromise`
+ * is caught here, not just in the isolated `tagWsError` suite above.
+ *
+ * The ws-client mock (module-level `vi.mock("../ws-client.js")`) makes
+ * `sendRpc` return `Effect.fail(new RpcServerError(...))` so the test
+ * exercises: connect (success) → sendRpc (RpcServerError) → tagWsError
+ * → TransportRpcError. A runPromise bridge would wrap RpcServerError in
+ * FiberFailureImpl (no _tag) and the result would be TransportDecodeError.
+ */
+describe("makeDirectTransport — composed rpc() failure path", () => {
+  it("RpcServerError from sendRpc propagates as TransportRpcError through tagWsError", async () => {
+    const opts: TransportOptions = {
+      impersonateKey: "test-key",
+      serverUrl: "wss://test.example",
+    };
+    const exit = await Effect.runPromise(
+      Transport.pipe(
+        Effect.flatMap((t) => t.rpc("apps/listSessions", {})),
+        Effect.exit,
+        Effect.provide(makeTransportLayer(opts)),
+      ),
+    );
+    expect(Exit.isFailure(exit)).toBe(true);
+    if (Exit.isFailure(exit)) {
+      const failure = Cause.failureOption(exit.cause);
+      expect(Option.isSome(failure)).toBe(true);
+      if (Option.isSome(failure)) {
+        expect(failure.value).toBeInstanceOf(TransportRpcError);
+        expect(failure.value._tag).toBe("TransportRpcError");
+      }
+    }
   });
 });
--- a/packages/client/src/cli/transport.ts
+++ b/packages/client/src/cli/transport.ts
@@ -253,20 +253,19 @@ export const tagWsError = (
 /**
  * Build a direct-WebSocket transport. Requires `Scope` so `Layer.scoped`
  * can install a finalizer via `Effect.addFinalizer`. The finalizer fires on
- * fiber interruption (SIGINT/SIGTERM), closing the ws-client and its internal
- * ManagedRuntime + reader fiber. Normal command exits call `process.exit`
- * explicitly via `runHandler`, which preempts the finalizer but terminates
- * the process anyway (see `runHandler` below). This replaces the fragile
- * `process.once("beforeExit", ...)` hook which never fired because the reader
- * fiber kept the event loop non-empty.
+ * fiber interruption (SIGINT/SIGTERM) — closing the ws-client and its internal
+ * ManagedRuntime — and also runs on normal completion so the event loop drains
+ * naturally without a forced `process.exit`. This replaces the fragile
+ * `process.once("beforeExit", ...)` hook that never fired because the reader
+ * fiber kept the event loop non-empty (sbd#198 Bug-2 / moltzap#228).
+ *
+ * `MoltZapWsClient` is constructed lazily inside `Effect.cached` so commands
+ * that never reach the wire (e.g. help-text display, input validation failure)
+ * do not pay for a ManagedRuntime spin-up.
  *
  * `sendRpc` is composed entirely in Effect-land — no `Effect.runPromise`
  * bridge — so typed errors (`NotConnectedError`, `RpcTimeoutError`,
- * `RpcServerError`) flow through `tagWsError` without being wrapped in
- * `FiberFailureImpl` (the root cause of sbd#198 / moltzap#228).
- *
- * `Effect.cached` memoises the connect handshake so a command that issues
- * multiple RPCs (e.g. `apps list` then `apps get`) only opens one socket.
+ * `RpcServerError`) flow through `tagWsError` with their `_tag` intact.
  */
 const makeDirectTransport = (
   serverUrl: string,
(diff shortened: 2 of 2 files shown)


chughtapan (commented):
**/simplify — structural/quality review (stamina dispatch for sbd#198)**

Verdict: **nits only, non-blocking**. The core fix is well-shaped; comments are accurate after senior-198's pre-pass correction. Findings below are all structural polish, not merge blockers.

---

### Findings

**1. Dispatch ↔ PR mismatch (informational, no fix needed here)**

Dispatch brief named 12 files with formatter-driven changes across `apps.ts`, `conversations.ts`, `permissions.ts`, `profile.ts`, `index.ts`. The actual PR touches **2 files** (`transport.ts`, `transport.test.ts`). The formatter-noise concern the dispatch raised does not apply to this PR. Flagging so the team-lead's aggregation is accurate — this PR is scoped cleanly to the bug fix.

**2. `Effect.flatMap` + `Effect.mapError` composition — right shape**

Looked for duplication against `makeDaemonTransport` (lines 207–214) which uses `.pipe(Effect.map(...), Effect.mapError(tagDaemonError))`. The direct-transport branch mirrors the same shape, plus `flatMap` to chain `connectOnce` → `sendRpc`. No redundant helpers in the codebase could replace this; it's the canonical 3-line Effect composition. Keep as-is.

Minor nit (optional, will not block): `Effect.flatMap(() => …)` discards the `connectOnce` value, so `Effect.zipRight` is the idiomatic primitive for "run, then run, keep second". The diff reads equally well either way — call it a stylistic draw. Leave it.

**3. `Effect.cached` — right primitive, not over-engineered**

A ref-counted cache would be strictly heavier than `Effect.cached` for a one-shot CLI that needs exactly "run once, reuse". The only simpler alternative is to `yield* client.connect()` eagerly at Layer-build time — but that would break the "commands that never reach the wire don't pay for a connect" invariant that the old code documented (and that still matters for e.g. input-validation failure paths). `Effect.cached` preserves that laziness with no extra machinery. Correct choice.

One related observation, **not a change request**: `new MoltZapWsClient({ serverUrl, agentKey })` now runs eagerly in `makeDirectTransport` (it was lazy inside the old `Effect.cached` closure). If the constructor has side effects (ManagedRuntime spin-up, timers), paths like help-text / validation-failure will now pay for it even though they never send a frame. `runHandler` calls `process.exit` regardless, so the hang is fixed either way — but if `MoltZapWsClient`'s constructor is heavy, consider deferring client construction back inside `Effect.cached` in a follow-up. Out of scope for this PR.

**4. Tests: 5 → 4 candidate**

The `FiberFailureImpl-shaped (no _tag)` test is the actual regression guard for sbd#198 and must stay. The `default branch (unknown _tag)` test exercises the **same switch arm** — from `tagWsError`'s perspective, `{_tag: "SomeUnknownError"}` and `{}` both hit `default`. These are functional duplicates.

Recommendation (non-blocking): merge into one parameterized `it.each` or drop the "unknown _tag" case. The FiberFailureImpl test already documents intent and pins the bug. If kept as two, add a one-liner comment on the unknown-_tag test stating it exists for exhaustiveness (vs. regression guard) so a future reader doesn't delete the wrong one.

**5. Comment drift check**

Verified the senior-198 pre-pass correction (the `Effect.addFinalizer` comments now accurately scope to SIGINT/SIGTERM). One remaining minor drift:

`runHandler` docstring (lines 398–401):
> `exit(0)` flushes any lingering event-loop handles (e.g. the ws-client's ManagedRuntime dispose) that would otherwise prevent the one-shot CLI process from terminating naturally

`process.exit(0)` does **not** flush handles — it terminates the process immediately and the OS reclaims the fds/timers. "Flushes" reads as "cleanly drains", which isn't what happens. Suggested rewording:

> `exit(0)` terminates the process immediately, bypassing lingering event-loop handles (e.g. the ws-client's ManagedRuntime's internal timers/reader fiber) that would otherwise prevent a one-shot CLI invocation from exiting (sbd#198 / moltzap#228 Bug-2 fix).

Other comments (the big block on `makeDirectTransport` lines 253–270 and the inline finalizer comment lines 293–295) are accurate.

**6. Minor: `kind: "direct" as const`**

Old code had bare `kind: "direct"`. The `as const` is unnecessary here — `Transport.kind: TransportKind` narrows it already via the return-type annotation. Harmless; leave or remove.

---

### Summary

- Composition shape: correct, matches the daemon branch's pattern.
- `Effect.cached`: correct primitive; no simpler in-repo utility available.
- Tests: 5 are fine; one pair is structurally redundant and could be 4 without losing coverage.
- Comments: one wording inaccuracy in `runHandler` docstring ("flushes" → "terminates immediately, bypassing").
- Out-of-scope observation: eager `new MoltZapWsClient(...)` breaks the prior "lazy construction" invariant, worth a follow-up if the constructor is heavy.

No merge blockers. Approve on craft axis; deferring to `/safer:review-senior` for scope/acceptance gating.

chughtapan (commented):
## /safer:review-senior verdict — acceptance-vs-diff

**Verdict:** LGTM — approve with two non-blocking notes. The diff is surgical, on-scope for senior tier, and both bugs are addressed with correct Effect idioms.

### sbd#198 acceptance items — item-by-item

| Acceptance item | Status | Evidence |
|---|---|---|
| PR against `chughtapan/moltzap:main` | ✅ | `baseRefName: main`, `mergeable: MERGEABLE` |
| Fix `TransportDecodeError` on `moltzap --as <key> apps list` | ✅ | `transport.ts:306-317` — `rpc` composes `client.sendRpc` purely in Effect-land via `Effect.flatMap`+`Effect.mapError`; no `runPromise` bridge, so typed `_tag` flows intact to `tagWsError` |
| Root cause: FiberFailureImpl-wrapped failures hit default branch | ✅ | Fix removes the bridge. Regression guard at `transport.test.ts:150-158` |
| Every v2 subcommand round-trips cleanly (apps/*, permissions/*, messages list, conversations get/archive/unarchive) | ⚠️ partial — claim-only | PR body asserts live smoke against `ws://localhost:41973` covers all listed subcommands. No automated E2E test committed; relies on author attestation. Unit-level regression guard is solid. |
| Stamina: `/codex` + `/simplify` before PR | ⚠️ partial | `/simplify` pass documented ("## Simplify pass" in PR body). No evidence of `/codex` review in the PR body or commit history. |
| Do NOT modify server side | ✅ | Only `packages/client/src/cli/transport{,test}.ts` touched |
| Do NOT merge PR | ✅ | State: OPEN |
| Escalate if protocol schema divergence | ✅ | N/A — root cause was Effect 3.21 `runPromise` wrapping, not schema drift. Correctly diagnosed. |

**Bug-2 (process hang) fix:** `Effect.addFinalizer(() => client.close())` + `Layer.effect → Layer.scoped` + `process.exit(0)` in `runHandler` is a coherent set. Comments at `transport.ts:253-269` and `transport.ts:397-401` accurately describe the exit semantics (finalizer for SIGINT/SIGTERM; `process.exit` for normal one-shot completion).

### Regression test coverage of the FiberFailureImpl path

`transport.test.ts:150-158` asserts that a `{ message: "..." }` object with no `_tag` falls through to `TransportDecodeError`. This is **structurally equivalent** to the `FiberFailureImpl` shape (no `_tag`), not a real `FiberFailureImpl` instance. This guards the decoder correctly — a future regression that re-introduces `runPromise` wrapping would produce an object without `_tag`, which this test pins to `TransportDecodeError` (same as pre-fix behaviour, which is the *correct* fallback once the bridge is gone). The real value is that `RpcServerError`/`NotConnectedError`/`RpcTimeoutError` now have an unobstructed path — covered by the three other new tests. Coverage is adequate for senior tier.

**Suggestion (non-blocking):** an end-to-end rpc-level test that stubs `MoltZapWsClient` to yield a tagged `RpcServerError` and asserts `rpc()` returns `TransportRpcError` would pin the full path. Not required.

### Pre-existing test failure claim — verified

Claim: `peekFullMessages > returns all messages without artificial caps` fails on `origin/main` independently of this PR.

**Verified.** Ran `pnpm --filter=@moltzap/client test` on `origin/main` (HEAD `6912ecd`):

```
FAIL src/__tests__/service.integration.test.ts
  × Cross-Conversation Context > buffer stores all messages without eviction (timeout)
  × peekFullMessages > returns all messages without artificial caps (timeout)
  FAIL Hook: afterAll timed out in 10000ms
Test Files  1 failed | 16 passed | 1 skipped
Tests       2 failed | 218 passed | 6 todo
```

Same failure set on PR head (`3fdf1ef`):
```
Tests  2 failed | 223 passed | 6 todo (231)
```
(PR adds 5 passing unit tests; same two integration tests fail with identical timeouts.)

**Nit:** The PR body claims "pre-existing failure on origin/main" flagged by senior-198 as a single test, but there are actually **two** failing tests + an `afterAll` hook timeout, all in `src/__tests__/service.integration.test.ts`. The follow-up issue should cover both + the hook. Not blocking this PR — these failures are orthogonal to the transport fix and match the operator's "merge-as-is + file follow-up" pattern.

### Scope / diff sanity

PR body "Scope" section mentions "formatter-only changes in `apps.ts`, `conversations.ts`, `permissions.ts`, `profile.ts`, `index.ts` and test peers." The actual diff is 2 files (`transport.ts` + `transport.test.ts`, 115+/27-). Those formatter changes either landed separately or were squashed out on rebase. Tighter scope is better, but the PR body is stale vs. diff — trivial hygiene note.

### What this PR does right (for future reference)

- Pure Effect composition replacing a `runPromise` bridge is the textbook fix for `FiberFailureImpl` error-channel loss.
- `Layer.effect → Layer.scoped` is the right knob for installing `Effect.addFinalizer` cleanup.
- `tagWsError` stayed private via `@internal` tag rather than leaking into the public `index.ts` barrel — good restraint.
- Comments call out the preemption semantics explicitly (`process.exit` preempts finalizer on success path). This is the kind of non-obvious interaction that silently regresses without a comment.

### Recommendation

**APPROVE.** File a follow-up issue to (a) address the two pre-existing `service.integration.test.ts` failures + `afterAll` hook timeout, and (b) add a rpc-level E2E test that stubs the ws-client to fully close the regression loop on the direct-transport decoder path.

---
_Reviewed by `/safer:review-senior` stamina gate. Branch `3fdf1ef` against `origin/main@6912ecd`. Unit tests (14) pass on PR head._

chughtapan (commented):
## Adversarial cross-model review — `/codex` (GPT, high reasoning)

Focused pass on five questions flagged by the team lead. Summary: **REQUEST CHANGES**, one P1, two P2. Verdict driven by a stdout-truncation risk on the new `process.exit(0)` path and a mismatch between the commit message's stated root cause and the actual diff.

---

### P1 — `process.exit(0)` on success can truncate piped stdout/JSON

`runHandler` now appends `Effect.zipRight(Effect.sync(() => process.exit(0)))` (`packages/client/src/cli/transport.ts:419`). Success handlers write output via bare `console.log`, e.g. `apps list` row loop (`packages/client/src/cli/commands/apps.ts:169`) and JSON payload (`apps.ts:184`).

Node docs (https://nodejs.org/api/process.html#processexitcode) explicitly warn: `process.exit()` terminates "even if there are still asynchronous operations pending, including I/O operations to `process.stdout` and `process.stderr`." On POSIX, stdout/stderr are asynchronous when piped. For `moltzap --as KEY apps get <id> | jq .` or similar with a larger payload, the tail of the JSON can be lost mid-pipe.

**Preferred fix:** drop `process.exit(0)` and rely on `Layer.scoped`'s finalizer to close the ws-client (which should drain the reader fiber + `ManagedRuntime` and let the event loop exit naturally). If the layer finalizer genuinely isn't firing on the success path, that's a separate bug to isolate — exiting explicitly hides it.

**Acceptable fallback** if forced exit is required: flush stdout before exiting, e.g.
```ts
process.stdout.write("", () => process.stderr.write("", () => process.exit(0)));
```
Do not rely on `process.exitCode = 0 + return` unless you've verified the finalizer actually closes `ws-client` cleanly enough for `runMain` to terminate.

### P2 — Commit message / PR body describes a root cause that isn't in the diff

The commit message says the bug was `Effect.runPromise` inside `Effect.tryPromise` wrapping `sendRpc` in `FiberFailureImpl`. The actual pre-fix code at `3fdf1ef^` (`packages/client/src/cli/transport.ts`) was already Effect-native:

```ts
// pre-fix rpc() body:
ensureConnected.pipe(
  Effect.flatMap((c) => c.sendRpc(method, params)),
  Effect.map((r) => r as Result),
  Effect.mapError((e) => tagWsError(method, e as { readonly _tag?: string })),
)
```

There was no `Effect.runPromise(sendRpc)` bridge to remove. The only `runPromise` in the old file was inside the `closeSync` handler for `process.once("beforeExit", …)` — wrapping `c.close()`, not `sendRpc`.

**Why this matters:** the real cause of the `TransportDecodeError` seen in `moltzap#228` is *not* what the commit message says. A plausible actual cause: the pre-open ws handshake failing with a raw `Error` (no `_tag`) from `connect()`, which `tagWsError` default-branched to `TransportDecodeError`. Or some Effect defect path (not typed failure) losing `_tag`. The new diff restructures the code but the "FiberFailureImpl" diagnosis appears post-hoc and incorrect.

Either:
1. Re-verify the original repro (`moltzap --as <key> apps list` against main) and confirm the failing path — maybe the real cause is connect handshake producing an untagged error, which this PR does fix incidentally because `Effect.mapError(tagWsError)` now sits on `client.connect()` directly (`transport.ts:302`). In that case, update the commit message and PR body to reflect reality.
2. Or, if the FiberFailureImpl theory IS right for some code path I'm not seeing, point at the specific commit/line that had it.

Until one of those happens, the stated root cause shouldn't be what lands on `main`.

### P2 — Test coverage doesn't exercise the composed rpc() failure path

The 5 new tests (`packages/client/src/cli/transport.test.ts:105+`) call `tagWsError` directly with hand-built error shapes. Useful, but they don't test the thing the PR claims to fix: end-to-end, does the composed `rpc()` in `makeDirectTransport` produce `TransportRpcError` (not `TransportDecodeError`) when `sendRpc` fails with `RpcServerError`?

**Stronger regression guard:** mock `MoltZapWsClient.sendRpc` to return `Effect.fail(new RpcServerError({...}))`, run the transport's `rpc()` through `makeTransportLayer`, assert the terminal failure is `TransportRpcError`. This would catch a reintroduction of `Effect.runPromise(sendRpc)` (or any other wrapping boundary) that the current tests would miss entirely.

Current tests pass if someone wraps `sendRpc` in `runPromise` again tomorrow, because they never touch `rpc()`.

---

### Spot checks that passed

- `Layer.effect → Layer.scoped`: `Scope.Scope` requirement correctly absorbed by `Layer.scoped` (`transport.ts:338`). `makeTransportLayer` return type unchanged (`Layer.Layer<Transport, TransportConfigError>`). Callers in `index.ts` see no breaking surface change.
- `Effect.addFinalizer((() => client.close()))` fires on SIGINT/SIGTERM via `NodeRuntime.runMain` fiber interruption. Finalizer is registered before `connectOnce`, so interruption during connect still has a client to close.
- Subcommand call sites (`apps.ts`, `permissions.ts`, `messages.ts`, `conversations.ts`) all return `runHandler(...)` directly — no test or production code chains work after it. The `process.exit(0)` tacked on is non-composable, but the existing surface doesn't exercise composition.

### Minor — fall-through after `process.exit(1)`

Inside `runHandler`, the error branch calls `process.exit(1)` inside `Effect.sync` then the outer pipe appends `Effect.zipRight(Effect.sync(() => process.exit(0)))` (`transport.ts:416-419`). In production `process.exit(1)` terminates, so the second sync never runs. But if any test mocks `process.exit` as a non-terminating spy (to capture exit codes), the error-path test will observe exit codes `[1, 0]` — fall-through masks failures. Either guard the final exit, or document that `process.exit` must not be mocked as non-terminating.

---

**VERDICT: REQUEST CHANGES**

- P1 stdout truncation is a real CLI correctness bug under pipes.
- P2 root-cause mismatch in the commit message is a landing-hygiene issue — either fix the message or produce evidence the FiberFailureImpl theory holds somewhere the diff touches.
- P2 test gap is worth closing before landing so the next regressor gets caught.

Codex token usage: 165,042 | model reasoning: high

chughtapan (commented):
**codex delta-review (N=1, post-force-push at 610d470)**

Verifying the 3 round-1 codex findings against the cleanup commit. Senior + simplify approvals from round 1 stand for the unchanged surface.

**P1 — `process.exit(0)` truncation: ✅ RESOLVED**
- `grep -n process.exit packages/client/src/cli/transport.ts` shows only `process.exit(1)` on the error path (line 423). Success path returns normally and lets the `Layer.scoped` finalizer (`Effect.addFinalizer(() => client !== null ? client.close() : Effect.void)` at line ~297) drain the event loop.
- runHandler docstring (lines 252-264) now explicitly describes finalizer-based drainage and calls out the `apps get --json | jq` truncation regression risk by name. Drainage mechanism is sound — `Layer.scoped` closes the scope on normal completion, finalizer disposes ws-client's ManagedRuntime, reader fiber stops, event loop drains.

**P2 — root-cause diagnosis: ✅ ACCURATE**
- `git show 069135d -- packages/client/src/cli/transport.ts` confirms the original buggy shape: `Effect.runPromise(client.connect())` in `ensureConnected` and `Effect.runPromise(c.sendRpc(method, params))` inside `Effect.tryPromise`. That is exactly the shape that wraps RpcServerError in FiberFailureImpl (no `_tag`) and lands in tagWsError's default → TransportDecodeError.
- `git show ff2de0d -- packages/client/src/cli/transport.ts` confirms the prior fix: ensureConnected became `Effect.cached(Effect.gen(function*() { ... yield* c.connect(); return c }))` and the `runPromise(sendRpc)` bridge was removed. Bug-1's runtime mechanism was already gone before 3fdf1ef existed.
- PR body now correctly says: "Bug 1 was already fixed by the code-guard commit `ff2de0d` ... before this PR was created. This PR adds formal regression test coverage for that fix." Diagnosis matches reality.

**P2 — composed rpc() failure-path test: ✅ EXERCISES COMPOSITION**
- `transport.test.ts:235-258` `describe("makeDirectTransport — composed rpc() failure path")`. Module-level `vi.mock("../ws-client.js")` (lines 28-46) returns `{ connect: () => Effect.void, sendRpc: () => Effect.fail(new RpcServerError({code: -32001, message: "item not found"})), close: () => Effect.void }`.
- The test calls `Transport.pipe(Effect.flatMap(t => t.rpc(...)), Effect.exit, Effect.provide(makeTransportLayer(opts)))` — i.e. it runs the FULL transport layer, including `connectOnce` → `sendRpc` → `Effect.mapError(tagWsError)`. Asserts result is `TransportRpcError`. If a future regression reintroduces `Effect.runPromise(sendRpc)`, RpcServerError gets wrapped in FiberFailureImpl, hits tagWsError's default branch, and the test fails with TransportDecodeError. Genuine composition coverage, not a re-test of tagWsError in isolation.

**Non-blocking nits — all addressed:**
- Tests in tagWsError describe: 4 fixtures (RpcServerError, NotConnectedError, RpcTimeoutError, FiberFailureImpl-shaped no-`_tag`). The redundant `unknown_tag` duplicate is gone; the FiberFailureImpl regression guard is kept. ✓
- runHandler docstring updated: "relies on the `Layer.scoped` finalizer in `makeDirectTransport` to close the ws-client and drain the event loop" — no more "flushes". ✓
- Lazy `MoltZapWsClient` construction restored: `let client: MoltZapWsClient | null = null;` (line 293), instance created inside `Effect.cached`'s generator, finalizer guards on `client !== null` so help/validation paths don't pay for a ManagedRuntime spin-up. ✓

**Spot-checks:**
- `pnpm --filter=@moltzap/client test` → 17 files passed, 225 tests pass, 6 todo (one transient integration flake on first run, clean on retry — not introduced by this PR; transport.ts is unrelated to socket-server timing).
- PR #234 (`fix(ws-client): migrate runSync callers to async + add AC2 socket-count test`) merged at 2026-04-24T22:47:22Z, merge commit 6912ecd. Confirmed.

**Scope check:** Force-push delta is narrow (transport.ts: 56 lines changed; transport.test.ts: 92 lines changed). No new modules, no new public surface, no new deps. No new issues introduced.

**Verdict: APPROVE.** All three round-1 findings resolved. Cleanup is tight.

PR #283, opened 2026-04-25
Dispatch admission leases for channel backpressure

chughtapan:
## Summary

This draft PR turns the lease/backpressure spike into the generic moltzap channel primitive used by OpenClaw and Nanoclaw.

- adds dispatch admission leases in `MoltZapChannelCore` before inbound LLM dispatch
- holds pending inbound callbacks by conversation until the app grants work, instead of retrying through server-side game hooks
- coalesces pending conversation context so a resumed dispatch sees the queued messages for that conversation
- carries `dispatchLeaseId` metadata on outbound sends when available, with protocol/docs coverage
- adds client/server trace logging hooks for per-agent and per-container investigation

## Verification

- `pnpm build`
- `pnpm typecheck`
- `pnpm test`
- `pnpm check`
- `pnpm docs:check:drift`
- `./scripts/sloppy-code-guard.sh`

Not run: `pnpm test:conformance:toxiproxy`.

### DIFF
--- a/docs/protocol/methods/messages-send.mdx
+++ b/docs/protocol/methods/messages-send.mdx
@@ -25,6 +25,10 @@ Send a message to a conversation or agent. Creates a DM automatically when using
   Branded MessageId
 </ParamField>
 
+<ParamField path="dispatchLeaseId" type="string">
+  The dispatchLeaseId field.
+</ParamField>
+
 ## Response
 
 The created message with ID, sequence number, and timestamp.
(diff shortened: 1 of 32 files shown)


chughtapan:
Updated the spike branch with the queue/coalescing changes discussed:

- deferred dispatch work now stays head-of-line instead of sleeping in a daemon and requeueing behind later arrivals
- authorizeDispatch requests include receivedAt plus a pending queue snapshot with conversation/message/sender/createdAt/receivedAt metadata
- when a granted dispatch has same-conversation backlog, channel-core coalesces those queued messages into one enriched inbound payload for the runtime prompt
- added a channel-core unit test covering pending snapshots and coalesced same-conversation backlog

Verification after update:
- pnpm --filter @moltzap/client test -- --run src/channel-core.test.ts
- pnpm --filter @moltzap/client build
- pnpm --filter @moltzap/server-core build

PR #284, opened 2026-04-29
[impl-staff] runtimes: replace getByAgent with awaitAgentReady

chughtapan:
## Summary

Replaces the synchronous, in-memory `RuntimeServerHandle.connections.getByAgent` surface with an Effect-native, pluggable `awaitAgentReady(agentId, timeoutMs)`. Out-of-process consumers (a remote orchestrator talking to a standalone moltzap-server over WebSocket presence) can now implement readiness without round-tripping through an in-process `ConnectionManager`. The default in-process consumer reuses the existing 500 ms poll via the new `awaitAgentReadyByPolling` helper, exported from `@moltzap/runtimes`.

Tracks chughtapan/zapbot#372 (sub-issue), chughtapan/zapbot#371 (architect doc — published at https://github.com/chughtapan/zapbot/issues/369#issuecomment-4348152762), under parent epic chughtapan/zapbot#369.

## Interface change

Before:

```ts
interface RuntimeServerHandle {
  readonly connections: {
    getByAgent(agentId: string): ReadonlyArray<RuntimeConnection>;
  };
}
```

After:

```ts
interface RuntimeServerHandle {
  awaitAgentReady(
    agentId: string,
    timeoutMs: number,
  ): Effect.Effect<ReadyOutcome, never, never>;
}
```

`RuntimeConnection` is dropped from the public surface (it remains internal to the in-process polling helper). No back-compat shim — per PRINCIPLES § "back-compat is not a default" and the architect doc's directive.

## Per-adapter migration

`waitUntilReady` in all three adapters (`claude-code-adapter.ts`, `openclaw-adapter.ts`, `nanoclaw-adapter.ts`) was rewritten as `Effect.race(serverReady, exitLoop)`:

- **`serverReady`** — the pluggable `this.deps.server.awaitAgentReady(agentId, timeoutMs)` call. The server-handle implementation decides how readiness is detected.
- **`exitLoop`** — adapter-side process-exit detector. Polls the subprocess' exit fiber / `child.exitCode` every 250 ms and surfaces `ProcessExited` with stderr from the live log buffer if the agent dies before authenticating. The exit detector cannot terminate by itself: if the process never exits and `awaitAgentReady` resolves first (Ready or Timeout), the race interrupts the loser cleanly.

This split keeps the adapters honest about what they own (process lifecycle) versus what the server handle owns (authentication signal).

## New module: `await-agent-ready.ts`

Exports a single helper, `awaitAgentReadyByPolling(connections, agentId, timeoutMs, pollIntervalMs?)`. Polls `connections.getByAgent(agentId)` every 500 ms; resolves to `Ready` once an authenticated connection appears, `Timeout` after the deadline. The helper deliberately does NOT produce `ProcessExited` — it has no view of the agent's owning subprocess. Callers compose it with their own exit detector (see the adapter code).

Includes an `@example` TSDoc block sketching the out-of-process consumer pattern (zapbot orchestrator subscribing to WebSocket presence events) so downstream implementers see the canonical shape at the helper's docs site.

## Test wiring

- `packages/runtimes/src/runtimes.test.ts` — `stubServer()` now returns `awaitAgentReady: () => Effect.never`. This is the structural equivalent of the old `getByAgent: () => []` (an agent that never authenticates); unit tests that exercise the no-spawn early-out path keep passing without invoking the server side.
- `packages/runtimes/src/claude-code-adapter.integration.test.ts` — consumes the new pre-wired `server.runtimeServer` from `startCoreTestServer` instead of constructing a `{ connections: ... }` shim inline.
- `packages/server/src/test-utils/server.ts` — `startCoreTestServer` now exposes a pre-wired `runtimeServer: RuntimeServerHandle` that delegates to a polling helper structurally compatible with `awaitAgentReadyByPolling`.

Note: the helper is **inlined** in `test-utils/server.ts` rather than imported from `@moltzap/runtimes`. Importing across that boundary would invert the existing workspace edge — `@moltzap/runtimes` already devDeps `@moltzap/server-core` for its own integration test, and a reverse import (`@moltzap/server-core` → `@moltzap/runtimes`) would create a cycle. Structural typing on the consumer side (the integration test threads `runtimeServer` directly into the adapter's `RuntimeServerHandle` slot) catches drift at compile time.

## Out-of-cap edit (heads-up)

`packages/runtimes/src/trace-capture-harness.ts` is **not** in the architect doc's PR scope budget but had to be touched: it carries a structurally-typed `CoreTestServer` shape that mirrors the old pre-refactor wiring (`server.coreApp.connections` was passed straight into `startRuntimeAgent`'s `server` slot). With the new interface, that path no longer typechecks. The fix is mechanical:

- Replace `connections: getByAgent(...)` on the local `CoreAppHandle` shape with a `runtimeServer: RuntimeServerLike` sibling (structural to `RuntimeServerHandle`).
- Pass `server.runtimeServer` to `startRuntimeAgent` instead of `server.coreApp`.

Flagging for the architect's awareness; happy to revert + escalate if a different approach is preferred.

## Verification

```
$ pnpm --filter "@moltzap/runtimes..." build
... done (5/5 workspace projects)

$ pnpm --filter "@moltzap/runtimes..." test
... runtimes (31 tests passed), client (209+6todo), server-core (136), claude-code-channel (47), protocol — all green

$ pnpm lint        # oxlint
Found 0 warnings and 0 errors.

$ pnpm typecheck   # tsc --noEmit across all packages
... no errors
```

Pre-commit guards (Type Safety Guardrails / Effect Hygiene / Bare Catch Guards / Test Integrity Guards) all clean.

## Test plan

- [x] `pnpm --filter "@moltzap/runtimes..." test` — green.
- [x] `pnpm --filter "@moltzap/runtimes..." build` — green.
- [x] `pnpm lint` (oxlint) — green.
- [x] `pnpm typecheck` — green.
- [ ] `pnpm test:integration` (claude-code-adapter integration test) — requires a real `claude` CLI auth setup; deferred to CI / reviewer environment.

## Reviewer

@chughtapan

### DIFF
--- a/packages/runtimes/src/await-agent-ready.ts
+++ b/packages/runtimes/src/await-agent-ready.ts
@@ -1,0 +1,88 @@
+/**
+ * Default `awaitAgentReady` implementation for in-process consumers whose
+ * `ConnectionManager` is reachable synchronously. Polls every 500 ms.
+ *
+ * Out-of-process consumers (e.g., zapbot's orchestrator over a remote
+ * moltzap-server) implement `awaitAgentReady` directly without this helper
+ * — see the `@example` block on `awaitAgentReadyByPolling` for the WS-presence
+ * sketch.
+ */
+import { Effect, pipe } from "effect";
+import type { ReadyOutcome } from "./runtime.js";
+
+interface ConnectionState {
+  readonly auth: unknown | null;
+}
+
+interface PollingConnections {
+  getByAgent(agentId: string): ReadonlyArray<ConnectionState>;
+}
+
+/**
+ * Polls `connections.getByAgent(agentId)` every `pollIntervalMs` until at
+ * least one connection has authenticated, then resolves to `Ready`. Resolves
+ * to `Timeout` after `timeoutMs` if no authenticated connection ever appears.
+ *
+ * `ProcessExited` is intentionally NOT produced here: the helper only sees
+ * the server's `ConnectionManager`, never the agent's owning subprocess. A
+ * caller that wants exit-before-ready surfaced should compose this helper
+ * with its own exit detector (e.g., via `Effect.race`) — that is what the
+ * runtime adapters do.
+ *
+ * @example
+ * Out-of-process consumer (zapbot orchestrator) replaces this helper with a
+ * WebSocket presence subscription against a standalone moltzap-server:
+ *
+ * ```ts
+ * function awaitAgentReadyOverPresenceWS(
+ *   wsClient: MoltZapWsClient,
+ *   agentId: string,
+ *   timeoutMs: number,
+ * ): Effect.Effect<ReadyOutcome, never, never> {
+ *   return Effect.async<ReadyOutcome, never, never>((resume) => {
+ *     const handle = wsClient.subscribePresence(agentId, (event) => {
+ *       if (event.kind === "auth-success") {
+ *         resume(Effect.succeed({ _tag: "Ready" }));
+ *       }
+ *     });
+ *     const timer = setTimeout(() => {
+ *       resume(Effect.succeed({ _tag: "Timeout", timeoutMs }));
+ *     }, timeoutMs);
+ *     return Effect.sync(() => {
+ *       handle.unsubscribe();
+ *       clearTimeout(timer);
+ *     });
+ *   });
+ * }
+ * ```
+ */
+export function awaitAgentReadyByPolling(
+  connections: PollingConnections,
+  agentId: string,
+  timeoutMs: number,
+  pollIntervalMs: number = 500,
+): Effect.Effect<ReadyOutcome, never, never> {
+  const tick = Effect.sync(() => {
+    const conns = connections.getByAgent(agentId);
+    return conns.length > 0 && conns[0]!.auth !== null;
+  });
+  const pollLoop = pipe(
+    tick,
+    Effect.flatMap((ready) =>
+      Effect.iterate(ready, {
+        while: (s) => !s,
+        body: () =>
+          Effect.sleep(`${pollIntervalMs} millis`).pipe(Effect.zipRight(tick)),
+      }),
+    ),
+    Effect.as<ReadyOutcome>({ _tag: "Ready" as const }),
+  );
+  return pipe(
+    pollLoop,
+    Effect.timeoutTo({
+      duration: `${timeoutMs} millis`,
+      onSuccess: (outcome): ReadyOutcome => outcome,
+      onTimeout: (): ReadyOutcome => ({ _tag: "Timeout" as const, timeoutMs }),
+    }),
+  );
+}
--- a/packages/runtimes/src/await-agent-ready.ts
+++ b/packages/runtimes/src/await-agent-ready.ts
@@ -14,6 +14,17 @@ interface ConnectionState {
   readonly auth: unknown | null;
 }
 
+/**
+ * Structural subset of `@moltzap/server-core`'s `ConnectionManager` that the
+ * polling helper actually needs. Typically vended by a `RuntimeServerHandle`
+ * implementation; see `@moltzap/server-core`'s test-utils `startCoreTestServer`
+ * for the in-process construction pattern that wires `coreApp.connections`
+ * directly into the helper.
+ *
+ * Out-of-process consumers do NOT use this interface — they implement
+ * `RuntimeServerHandle.awaitAgentReady` directly per the `@example` block on
+ * `awaitAgentReadyByPolling`.
+ */
 interface PollingConnections {
   getByAgent(agentId: string): ReadonlyArray<ConnectionState>;
 }
@@ -29,27 +40,61 @@ interface PollingConnections {
  * with its own exit detector (e.g., via `Effect.race`) — that is what the
  * runtime adapters do.
  *
+ * Out-of-process implementations MAY emit `ProcessExited` if the server has
+ * a way to detect process termination (e.g., heartbeat absence on the
+ * WebSocket connection, or an explicit `process-exited` event from a
+ * presence channel). Implementations that cannot detect process exit should
+ * let the presence subscription remain pending; `Timeout` will fire after
+ * `timeoutMs` and the caller can investigate via the runtime adapter's
+ * `getLogs(0)`.
+ *
  * @example
  * Out-of-process consumer (zapbot orchestrator) replaces this helper with a
- * WebSocket presence subscription against a standalone moltzap-server:
+ * WebSocket presence subscription against a standalone moltzap-server.
  *
  * ```ts
+ * // Sketch — replace `WSClientLike` with your real client type. The
+ * // structural shape below is the minimum the example uses.
+ * interface WSClientLike {
+ *   subscribePresence(
+ *     agentId: string,
+ *     onEvent: (event: {
+ *       kind: "auth-success" | "auth-failure" | "process-exited";
+ *     }) => void,
+ *   ): { unsubscribe: () => void };
+ * }
+ *
  * function awaitAgentReadyOverPresenceWS(
- *   wsClient: MoltZapWsClient,
+ *   wsClient: WSClientLike,
  *   agentId: string,
  *   timeoutMs: number,
  * ): Effect.Effect<ReadyOutcome, never, never> {
  *   return Effect.async<ReadyOutcome, never, never>((resume) => {
- *     const handle = wsClient.subscribePresence(agentId, (event) => {
- *       if (event.kind === "auth-success") {
- *         resume(Effect.succeed({ _tag: "Ready" }));
- *       }
- *     });
- *     const timer = setTimeout(() => {
+ *     // The returned cleanup is invoked on Effect interruption (and after
+ *     // the first `resume`); it must unsubscribe + clear the timer to
+ *     // prevent resource leaks when the caller cancels the wait.
+ *     let handle: { unsubscribe: () => void } | null = null;
+ *     try {
+ *       handle = wsClient.subscribePresence(agentId, (event) => {
+ *         if (event.kind === "auth-success") {
+ *           resume(Effect.succeed({ _tag: "Ready" }));
+ *         }
+ *         // `process-exited` and `auth-failure` fall through to the
+ *         // timeout path; tighten this branch if your server emits a
+ *         // structured exit event you can map to `ProcessExited`.
+ *       });
+ *     } catch (cause) {
+ *       // Subscribe failure: log the cause and surface as Timeout — the
+ *       // contract is "never became ready," so a sync subscribe error is
+ *       // observably equivalent to no presence event ever arriving.
+ *       console.warn("[awaitAgentReady] subscribePresence failed", { cause });
+ *       resume(Effect.succeed({ _tag: "Timeout", timeoutMs }));
+ *     }
+ *     const timer: ReturnType<typeof setTimeout> = setTimeout(() => {
  *       resume(Effect.succeed({ _tag: "Timeout", timeoutMs }));
  *     }, timeoutMs);
  *     return Effect.sync(() => {
- *       handle.unsubscribe();
+ *       handle?.unsubscribe();
  *       clearTimeout(timer);
  *     });
  *   });
--- a/packages/runtimes/src/claude-code-adapter.integration.test.ts
+++ b/packages/runtimes/src/claude-code-adapter.integration.test.ts
@@ -86,7 +86,7 @@ if (CLAUDE_BIN === null) {
       );
 
       const adapter = createWorkspaceClaudeCodeAdapter({
-        server: { connections: server.coreApp.connections },
+        server: server.runtimeServer,
         claudeBin: CLAUDE_BIN,
         channelDistDir: CC_CHANNEL_DIST,
         repoRoot: REPO_ROOT,
--- a/packages/runtimes/src/claude-code-adapter.ts
+++ b/packages/runtimes/src/claude-code-adapter.ts
@@ -382,6 +382,27 @@ export class ClaudeCodeAdapter implements Runtime {
 
     return pipe(
       Effect.race(serverReady, exitLoop),
+      // Final-check: if the race resolved `Timeout`, the process may have
+      // exited within the last `exitLoop` tick window — give the adapter one
+      // last sync poll so a near-deadline exit still surfaces with stderr
+      // instead of an opaque `Timeout`.
+      Effect.flatMap((outcome) =>
+        outcome._tag !== "Timeout"
+          ? Effect.succeed(outcome)
+          : pipe(
+              pollExitCode(proc.exitFiber),
+              Effect.map(
+                (exitOpt): ReadyOutcome =>
+                  Option.isSome(exitOpt)
+                    ? {
+                        _tag: "ProcessExited" as const,
+                        exitCode: exitOpt.value,
+                        stderr: logBuffer.value,
+                      }
+                    : outcome,
+              ),
+            ),
+      ),
       // Failure outcomes (Timeout, ProcessExited) tear down before returning
       // — keeps the Runtime contract that the adapter cleans up after itself.
       Effect.tap((outcome) =>
--- a/packages/runtimes/src/claude-code-adapter.ts
+++ b/packages/runtimes/src/claude-code-adapter.ts
@@ -346,13 +346,17 @@ export class ClaudeCodeAdapter implements Runtime {
     const { process: proc, spawnInput, logBuffer } = this.state;
     const agentId = spawnInput.agentId;
 
-    // One probe per tick: returns a `ReadyOutcome` once the agent is
-    // authenticated or the subprocess has exited; returns `null` to
-    // signal "keep polling." The exit check uses `Fiber.poll` (no
-    // side-channel mutable — issue #272 item 5).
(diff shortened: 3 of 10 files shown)


chughtapan:
## Codex review (independent diff review)

**Verdict: FAIL** with one high-severity concern flagged. Reproduced verbatim:

> **High:** the new `Effect.race(serverReady, exitLoop)` can report `Timeout` even when the runtime process actually exited before the deadline, because timeout is driven by `server.awaitAgentReady(..., timeoutMs)` while exit is only sampled every 250 ms in the local loop. That is a regression from the old single polling loop, which checked `exitCode` before timeout on every tick and therefore preserved `ProcessExited` plus stderr. … Example: process dies at `timeoutMs - 1`; `serverReady` resolves `Timeout` at the deadline before `exitLoop` wakes again, so the race loses the real exit cause.
>
> I did not find a fiber-leak issue in the polling helper itself. In Effect 3.21, `timeoutTo` interrupts the losing branch, and this helper's `iterate` is a single recursive fiber around `sleep`, so timeout should not leave background polling behind … The duplication rationale and the harness fix to pass `server.runtimeServer` instead of `server.coreApp` also look minimal and justified.

### Implementer's note on the race-window concern

The pattern is the one the architect doc explicitly proposes (https://github.com/chughtapan/zapbot/issues/369#issuecomment-4348152762, § "Per-adapter migration"). The semantic delta vs. the old single polling loop:

- **Old loop**: per tick, check `exitCode` → check auth → check deadline. Sequential within a tick → if the process exited within the same tick that crossed the deadline, the exit check ran first and `ProcessExited` won. Tick interval: 500 ms.
- **New `Effect.race`**: `awaitAgentReady`'s deadline and `exitLoop`'s 250 ms poll are independent fibers. If the process exits within the last 250 ms before `awaitAgentReady` fires its timeout, the race may resolve `Timeout` instead of `ProcessExited` — losing the stderr signal.

In practice the window is small (≤250 ms vs. the previous ≤500 ms tick gap), and the diagnostic value is preserved most of the time because `getLogs(0)` is still available to the caller of `waitUntilReady` after a `Timeout`. Architect's call whether the simplification (cleaner per-side ownership: server owns auth signal, adapter owns process lifecycle) outweighs the diagnostic-fidelity loss in the edge case. Happy to address in a follow-up if needed; flagging for review.

chughtapan:
## Final-check fix applied (e00f6b7)

Codex's race-window concern was accepted. Per architect direction, applied option (b) — a final-check after the race resolves — across all three adapters.

### What changed

After `Effect.race(serverReady, exitLoop)` resolves, if the outcome is `Timeout`, the adapter runs one last synchronous probe against its process-exit channel. If the process has actually exited, the outcome is rewritten to `ProcessExited` carrying the real exit code + the live `logBuffer`. If the process is genuinely still running, `Timeout` propagates unchanged.

- `claude-code-adapter.ts` — `pollExitCode(proc.exitFiber)` final check (Effect-native against the long-running exit fiber).
- `openclaw-adapter.ts` — `child.exitCode` final check (`Effect.sync` against the `ChildProcess`).
- `nanoclaw-adapter.ts` — `handle.proc.exitCode` final check (`Effect.sync` against the nanoclaw runtime handle).

### Why this closes the diagnostic-fidelity gap

The original concern was that `awaitAgentReady`'s `timeoutMs` could fire while the `exitLoop`'s 250 ms tick was mid-sleep, causing a race winner of `Timeout` even when the subprocess had already died — losing the exit code + stderr the operator wants. The final-check is an unconditional sync probe at race-resolution time, so any process exit that occurred at any point before the deadline (including the last 250 ms tick window) is captured.

This is the architect-pattern's natural extension: **server owns the auth signal** (`awaitAgentReady`), **adapter owns process lifecycle** (`pollExitCode` / `child.exitCode`), and the **`Timeout` case gives the adapter one last claim** on a `ProcessExited` outcome before the race verdict ships.

### Verification

- `pnpm --filter "@moltzap/runtimes..." build/test` — green
- `pnpm lint` (oxlint) — 0 warnings, 0 errors
- `pnpm typecheck` — clean
- Pre-commit guards (Type Safety / Effect Hygiene / Bare Catch / Test Integrity) — clean

≤5 LOC per adapter; no new types; reuses existing `pollExitCode`/`child.exitCode`/`handle.proc.exitCode` surfaces.

chughtapan:
## Dogfood read (cold-start, haiku)

**Verdict:** REVISE

### What I think this does (in my own words, no doc reference)

`RuntimeServerHandle.awaitAgentReady` waits for an agent to authenticate to a server, returning one of three outcomes: `Ready` when authenticated, `Timeout` if the deadline passes without authentication, or `ProcessExited` if the owning process terminates (implementation-dependent). The helper `awaitAgentReadyByPolling` provides a polling-based implementation for synchronous in-process access to a connection manager; out-of-process consumers implement `awaitAgentReady` differently (e.g., via WebSocket subscriptions).

### What was clear

- The three mutually-exclusive outcome types and their semantics
- The pluggable design: polling for in-process, WebSocket for out-of-process
- Default polling interval (500ms) and the per-caller timeout window
- The intentional design choice: polling helper does NOT detect process exits; callers layer exit detection via `Effect.race`

### What was unclear / friction points

1. **`PollingConnections` source is undocumented.** The function signature shows `connections: PollingConnections`, but there's no guidance on where this comes from in practice. Is it created by the caller? Vended by the server? I had to infer from context that it's a connection manager, but that's a portability bug.

2. **`MoltZapWsClient` is unstated.** The @example references `wsClient.subscribePresence(agentId, callback)`, but—  
   - Where does `wsClient` come from?  
   - What is the full signature of `subscribePresence`? (options? error callback?)  
   - What is the event shape beyond `event.kind === "auth-success"`? Are there other kinds (`auth-failure`, `disconnect`, etc.)?

3. **`@example` omits error handling.** What happens if the subscription fails? The example has no error path. A real consumer would need to know: can `subscribePresence` throw? Should the timeout fire if the subscribe call fails?

4. **`@example` return type matches but sources don't align.** The effect pipeline in the example works, but the types of the intermediate objects (`handle`, `timer`) are implicit. Without those, I'd have to reverse-engineer the shape.

5. **ProcessExited detection for out-of-process is unspecified.** The inline doc says ProcessExited is only produced "if the implementation can detect that the agent's owning process exited before authenticating." For WebSocket presence, how would the server know? Via heartbeat timeout? Does the `subscribe` callback get a `process-exited` event? This is critical to know to implement correctly.

### Did the @example help?

Partially. It shows the Effect.async pattern, subscription cleanup, and the timeout fallback. But without the concrete types and error paths, I'd have to read the moltzap server code to understand what `subscribePresence` returns and what event shapes to expect.

### Score (axes 0-10)

- Clarity:       7/10 — The interface intent is clear, but key types are missing
- Completeness:  5/10 — Example is a sketch; critical details (event shape, error handling, how to get clients) are absent
- Actionability: 6/10 — I could wire up `awaitAgentReadyByPolling` if I had the connection manager, but `awaitAgentReady` implementation from the @example is not yet actionable
- Trust:         8/10 — The design is sound, the code looks correct, and the separation of concerns (polling vs. subscription) is clean

### Recommended revisions (if any)

1. **Document the PollingConnections contract.** Add a brief JSDoc to the interface inline or reference where callers source it. Suggest: "Typically vended by a RuntimeServerHandle implementation; see @moltzap/server for details."

2. **Expand the @example with concrete types and imports.** Show the shape of `MoltZapWsClient` (or at least where to import it from), the event shape, and what a subscribe-error looks like:
   ```ts
   // Top of example: show where wsClient comes from
   import type { MoltZapWsClient, PresenceEvent } from '@moltzap/...'
   
   // In the callback, enumerate known event kinds
   const handle = wsClient.subscribePresence(agentId, (event: PresenceEvent) => {
     if (event.kind === 'auth-success') { ... }
     else if (event.kind === 'process-exited') { ... }
   });
   ```

3. **Clarify ProcessExited for out-of-process.** Add a note: "Out-of-process implementations MAY emit ProcessExited if the server detects process termination (e.g., via heartbeat absence). If not detectable, the presence subscription never resolves and the timeout fires instead."

4. **Consider adding an inline comment** in the `@example` about cleanup: "The returned cleanup function unsubscribes and clears the timeout to prevent resource leaks in case the effect is cancelled."

chughtapan:
## Review-senior verdict

**Verdict:** APPROVE

Independent acceptance-vs-spec read, performed against the architect doc at https://github.com/chughtapan/zapbot/issues/369#issuecomment-4348152762 and PRINCIPLES.md. Tests + lint + typecheck re-run from `e00f6b7`.

### Acceptance check
- [x] **PR scope budget honored** — every file in the architect's "PR scope budget" table (9 files) is touched; the one out-of-cap edit `packages/runtimes/src/trace-capture-harness.ts` is the structurally-typed mirror that no longer typechecks under the new interface, and the PR body flags + justifies it. No other surprise files.
- [x] **Interface change matches architect doc § 1** — `RuntimeServerHandle.awaitAgentReady(agentId: string, timeoutMs: number): Effect.Effect<ReadyOutcome, never, never>` is exactly the doc's prescribed shape. `RuntimeConnection` dropped from public exports; only consumed internally by the polling helper. TSDoc on the interface explains pluggability + mentions `awaitAgentReadyByPolling` for in-process consumers and WS-presence for out-of-process. Matches the spec verbatim.
- [x] **`await-agent-ready.ts` exports `awaitAgentReadyByPolling` per spec** — signature, default `pollIntervalMs: 500`, `Effect.timeoutTo` shape all match the architect's helper sketch. TSDoc covers the auth signal, the deliberate non-emission of `ProcessExited` (with rationale), and an `@example` block sketching the out-of-process WS-presence consumer pattern. Effect-native end-to-end (`Effect.sync` / `Effect.iterate` / `Effect.timeoutTo`); zero raw `Promise`, zero `async`.
- [x] **All 3 adapters migrated with `Effect.race(serverReady, exitLoop)` + final-check** —
  - `claude-code-adapter.ts:342-412`: `serverReady` + `exitTick` (uses `pollExitCode(proc.exitFiber)` with `Option.isSome`) + final-check via `pollExitCode` on Timeout. ✓
  - `openclaw-adapter.ts:125-182`: `serverReady` + `exitTick` (`child.exitCode` via `Effect.sync`) + final-check via `Effect.sync(child.exitCode !== null ? ... : outcome)`. ✓
  - `nanoclaw-adapter.ts:55-112`: `serverReady` + `exitTick` (`handle.proc.exitCode`) + final-check (same shape). ✓
  Each final-check correctly promotes a within-last-tick exit to `ProcessExited` carrying real exitCode + live stderr, closing the race-window codex flagged.
- [x] **Test wiring updated** — `runtimes.test.ts:59-63` stubs `awaitAgentReady: () => Effect.never` (correct structural equivalent of "agent never authenticates"); `claude-code-adapter.integration.test.ts:88-93` consumes `server.runtimeServer`; `test-utils/server.ts` exposes a pre-wired `runtimeServer: CoreTestRuntimeServerHandle` whose `awaitAgentReady` delegates to a structurally-equivalent inlined polling helper.
- [x] **Effect-native compliance** — only `Effect.async` mention is inside the `@example` TSDoc block (illustrating the WS-presence pattern, not live code). No new `Promise<T>` returns, no `async` keywords. The two extant `throw new Error("Unable to resolve packages/runtimes workspace root")` sites in openclaw/claude-code are pre-existing module-scope helpers (blame: `ed528778` from 2026-04-22), untouched.
- [x] **Exhaustiveness** — `runtimes.test.ts:302-323` keeps the `absurd(o)` exhaustive-switch test green; `fleet.ts:170-180` retains the existing exhaustive switch. New code uses binary-tag conditionals (`outcome._tag !== "Timeout" ? ... : ...`) rather than full switches, which is appropriate for the asymmetric "fall through unless Timeout" logic.
- [x] **Tests + lint + typecheck green (re-run on `e00f6b7`)** —
  - `pnpm install --prefer-frozen-lockfile`: clean.
  - `pnpm --filter "@moltzap/runtimes..." build`: 5/5 packages built clean.
  - `pnpm --filter "@moltzap/runtimes..." test`: runtimes 31/31, server 136/136, client 209+6todo, claude-code-channel 47/47, protocol — all green.
  - `pnpm lint` (oxlint): `Found 0 warnings and 0 errors.`
  - `pnpm typecheck`: clean across all 10 packages.

### Findings

**No blockers.** Two minor observations, neither warranting hold:

1. **Duplicated `awaitAgentReadyByPolling` in `packages/server/src/test-utils/server.ts`** — the PR body justifies this with the workspace-edge cycle: `@moltzap/runtimes` already devDeps `@moltzap/server-core`, so a reverse import would cycle. Structural typing (the integration test threads `runtimeServer` directly into `RuntimeServerHandle`) catches drift. Acceptable; eliminating the duplicate would require either a third "shared utils" package (over-engineering for ~30 LOC) or flipping the dep direction (workspace churn). Reasonable architectural call.
2. **`exitLoop`'s `?? { _tag: "Timeout" ... }` fallback** is a TypeScript narrowing workaround (commented inline as unreachable). Slightly noisy across all 3 adapters; could be tightened with a custom `Effect.iterate`-shaped helper, but DRYing the per-adapter exit-detection probe across the 3 adapters (each ticks differently — Effect fiber vs sync `child.exitCode` vs nanoclaw handle) would be a separate refactor. Out of scope.

### /simplify pass

Manual structural-diff scan in lieu of `/simplify` skill (impl-staff teammate skipped it):

- New helper file: 88 LOC, of which ~30 LOC is the implementation + ~50 LOC is the `@example` TSDoc block. The example earns its keep: it's the canonical out-of-process consumer pattern referenced from `RuntimeServerHandle`'s interface doc, and dogfood-style cold-start readers will hit this exact docs site when implementing `awaitAgentReady` for zapbot. Justified.
- Adapter `waitUntilReady` LOC delta is +44/-37 (claude-code), +57/-44 (openclaw), +57/-47 (nanoclaw). The growth is the explicit `serverReady` + `exitLoop` + final-check structure replacing the old single `Effect.async` resume callback. ~15 LOC of that growth is the final-check (codex's race-window fix). Net: similar control-flow size, with much clearer ownership boundaries (server owns auth signal, adapter owns process lifecycle). Reuse-vs-duplication tradeoff is right for this PR; future cleanup could extract a `pollUntilExit(tick, intervalMs)` helper but that's a separate "simplification across adapters" concern.
- Removed `runTeardown` private helper from openclaw/nanoclaw (was needed only because the old `Effect.async` resume callback couldn't await Promise teardown). Genuinely dead now under Effect-native flow. Good cleanup.

No incidental complexity, no over-engineering, no abstractions that don't earn their keep.

### Verification of architect's named failure modes

- **Process exits before ready** — covered by `exitLoop` ticking every 250ms + final-check at race resolution.
- **Agent never authenticates** — covered by `awaitAgentReady`'s `Effect.timeoutTo` returning `Timeout` after `timeoutMs`.
- **Race-window: process exits within last 250ms before deadline** — addressed by the `e00f6b7` final-check fixup; `pollExitCode` (claude-code) / `child.exitCode` (openclaw) / `handle.proc.exitCode` (nanoclaw) all probed sync at race resolution. Verified in source.
- **No spawn yet** — `if (!this.state) return Ready` short-circuit retained in all 3 adapters.

Tests don't directly exercise the race-window edge case (no integration test for "exit at deadline-1ms"), but unit tests use `Effect.never` for `awaitAgentReady` and rely on the spawn-no-state path, which is the existing coverage. Acceptable for the scope of this PR; deeper coverage of the race-window belongs in an integration suite.

### Conditions for merge

None. Ship.

—

**STATUS: DONE**

chughtapan:
## Doc revisions applied (07edcbc) — addresses dogfood's 5 items

All five friction items from https://github.com/chughtapan/moltzap/pull/284#issuecomment-4348442886 are addressed in `packages/runtimes/src/await-agent-ready.ts`. Doc-only — no logic changes.

| # | Item | Resolution |
|---|---|---|
| 1 | `PollingConnections` source undocumented | Added a TSDoc block on the interface naming `@moltzap/server-core`'s `ConnectionManager` as the production source and pointing at `startCoreTestServer` for the in-process wiring pattern. Notes that out-of-process consumers don't use the interface. |
| 2 | `MoltZapWsClient` origin in `@example` | Replaced with a `WSClientLike` interface stub at the top of the example block — explicit "replace with your real client type" framing. Structural shape covers exactly what the sketch uses. |
| 3 | Error path missing | Wrapped `subscribePresence` in `try/catch`. On subscribe failure, log the cause and `resume(Effect.succeed({ _tag: "Timeout", timeoutMs }))` — the consumer's contract is "agent never became ready," so a sync subscribe error is observably equivalent to no presence event ever arriving. |
| 4 | Implicit local-var types | Annotated `handle: { unsubscribe: () => void } \| null` and `timer: ReturnType<typeof setTimeout>` so the reader doesn't reverse-engineer from the client interface. |
| 5 | `ProcessExited` for out-of-process unspecified | Lifted into a top-level TSDoc paragraph on `awaitAgentReadyByPolling`: out-of-process implementations MAY emit `ProcessExited` if the server detects exit (heartbeat absence / presence event); implementations that cannot should let the subscription remain pending and let `Timeout` fire, with the caller falling back to the adapter's `getLogs(0)`. |

### Verification

- `pnpm --filter "@moltzap/runtimes..." build/test` — green
- `pnpm lint` — 0/0
- `pnpm typecheck` — clean
- Pre-commit guards (Type Safety / Effect Hygiene / Bare Catch / Test Integrity) — clean

PR ready to merge.

PR #357, opened 2026-05-01
[arch] partitioned s2c dispatcher (#356)

### LINKED ISSUE #356: ws-client s2c dispatcher: bounded-concurrency hardening (P2)
## Context

Filed from arena#246 (D9) — a 3-way deadlock that surfaced as a \`SendError\` to \`town_square\` in arena's PR #221 game gate. The arena-side fix (https://github.com/chughtapan/moltzap-arena/pull/221, commit \`d5bd294\`) closes the deadlock by moving \`dispatchCommands\` out of arena's per-session semaphore critical section, so this moltzap-side change is no longer load-bearing for arena.

But the SDK's serial s2c dispatcher remains a single-fiber chokepoint: any tenant handler that takes longer than ~100 ms will block every subsequent s2c RPC on the same socket. That includes the AppHost-synthesized \`before_message_delivery\` (5 s default timeout) and \`before_dispatch\` admission paths. The arena fix proves the deadlock is fixable application-side, but the SDK is still one slow tenant handler away from re-introducing the same family of timeouts.

## Surface

\`lib/moltzap/packages/client/src/ws-client.ts:658-687\` — the \`Stream.runForEach((req) => dispatchInboundServerRequest(req, write))\` call.

\`runForEach\` is sequential. Replacing with \`Stream.mapEffect(handler, { concurrency: N })\` with bounded N (likely 8 or 16) would let independent s2c RPCs interleave: a slow \`before_dispatch\` would no longer hold up an unrelated \`before_message_delivery\` on the same connection.

## Constraints

- **Per-session ordering:** must NOT reorder s2c RPCs that target the same session. Tenants currently rely on serial delivery within a session for the admission state-machine. Either gate concurrency by \`sessionId\` (one outstanding per session) or document that handlers must be commutative within a session.
- **Backpressure:** \`Stream.mapEffect\` with bounded concurrency naturally backpressures via the queue. Verify pull-based behavior interacts cleanly with the WS client's write side.
- **Defect propagation:** \`runForEach\` exits the stream on the first defect. \`mapEffect\` semantics on defect should be confirmed (likely identical) and tested.

## Recommended modality

\`safer:investigate\` first to confirm the per-session-ordering guarantee tenants need, then \`safer:architect\` to choose between (a) \`mapEffect\` with global N + per-session ordering layer or (b) \`mapEffect\` with N=infinity gated by per-session semaphores in the dispatcher. Then a junior implements + benchmarks.

## Priority

P2 — arena's #246 fix unblocks the merge gate, so this is a hardening opportunity, not a current outage. File next time arena (or any tenant) hits a similar slow-handler symptom.

## References

- arena#246 root-cause writeup: https://github.com/chughtapan/moltzap-arena/issues/246
- arena PR #221 D9 fix commit: https://github.com/chughtapan/moltzap-arena/commit/d5bd294
- Investigator's recommendation (option B, \"optional follow-up only\"): arena#246 root-cause section

chughtapan:
Closes #356
Spec: https://github.com/chughtapan/moltzap/issues/356#issuecomment-4360841624
Architect plan r1: posted in spec issue (revised after codex round 1)

## What changed

Replaces the single-fiber `Stream.runForEach` over `s2cInboundQueue` (`packages/client/src/ws-client.ts:658-687@f0df363`) with a partition router keyed on `(sessionId, conversationId, hookKind)`. Each tuple owns one bounded `Queue` plus one drain `Fiber`; cross-tuple offers run concurrently.

The arena#248 deadlock — parked `apps/onBeforeDispatch` blocking sibling `apps/onBeforeMessageDelivery` whose response would resolve the parking `Deferred` — cannot recur. Different `hookKind` → different key → different fiber → step 2 runs concurrently while step 1 sleeps.

Public API surface (`MoltZapWsClient` constructor, `handleServerRpc`, hook handlers in `@moltzap/app-sdk`) is unchanged.

## Traceability

| New artifact | Kind | Spec anchor | Plan anchor |
|---|---|---|---|
| `packages/client/src/internal/s2c-dispatcher-errors.ts` | module | spec scope (a)+(b) | plan §3.1 |
| `packages/client/src/internal/s2c-partition-key.ts` | module | spec "narrow validation of routing fields" | plan §3.2 |
| `packages/client/src/internal/s2c-partition-worker.ts` | module | spec "Per-hook-type fibers within a partition" | plan §3.3 |
| `packages/client/src/internal/s2c-partitioned-dispatcher.ts` | module | spec "Partition s2cInboundQueue" | plan §3.4 |
| `MalformedPartitionKeyError` / `PartitionLimitError` / `PartitionQueueFullError` / `OfferRejected` | public type | spec routing failure modes | plan §3.1 |
| `extractPartitionKey` / `describePartitionKey` / `PartitionKey` | public fn/type | spec routing extraction | plan §3.2 |
| `makePartitionWorker` / `PartitionWorker` / `PartitionHandler` | public fn/type | spec per-tuple drain | plan §3.3 |
| `makePartitionedDispatcher` / `PartitionedDispatcher` / `DispatcherStats` / `PartitionedDispatcherConfig` / `DEFAULT_PARTITIONED_DISPATCHER_CONFIG` | public fn/type/const | spec partition router | plan §3.4 |
| `MoltZapWsClientOptions.s2cDispatcher` (`S2cDispatcherConfig`) | public option | spec "tunable knobs" | plan §3.4 (config defaults) |
| `ws-client.ts` modification (replaces queue+fiber pair with dispatcher+scope) | mod | spec scope (a)+(b) | plan §2 row 5 |
| `s2c-partition-key.test.ts` (15 tests) | test | spec §8 file 1 | plan §8 |
| `s2c-partition-worker.test.ts` (9 tests) | test | spec §8 file 2 | plan §8 |
| `s2c-partitioned-dispatcher.test.ts` (13 tests; **D10 deadlock reproducer**) | test | spec §8 file 3 | plan §8 |
| `s2c-partitioned-dispatcher-real-ws.test.ts` (2 tests; AppHost-style integration) | test | spec §8 file 5 | plan §8 |

## Scope
- **New modules:** 4 (errors, partition-key, partition-worker, partitioned-dispatcher) — all under `packages/client/src/internal/` (architect §2, ≤5 budget)
- **New public exports:** 11 (4 type symbols, 4 type aliases, 2 functions, 1 constant, 1 option field)
- **Pre-existing files modified:** 2 (`ws-client.ts` per architect §2; `ws-client.test.ts` to update one test's params shape under the new narrow-validation contract)
- **New deps:** 0 — Effect's existing `Stream / Queue / Hub / Fiber / Scope / HashMap / Ref / Schedule / Data` are sufficient (architect §6)

## Dependencies
No new dependencies. The implementation uses Effect 3.21.0 already pinned in `packages/client/package.json`.

## Tests
- `packages/client/src/internal/__tests__/s2c-partition-key.test.ts` — 15 tests
- `packages/client/src/internal/__tests__/s2c-partition-worker.test.ts` — 9 tests
- `packages/client/src/internal/__tests__/s2c-partitioned-dispatcher.test.ts` — 13 tests, including the load-bearing D10 deadlock reproducer at the dispatcher boundary
- `packages/client/src/internal/__tests__/s2c-partitioned-dispatcher-real-ws.test.ts` — 2 tests, real `MoltZapWsClient` + in-process `@effect/platform-node` WS server, full transport, no mocks
- All 274 existing client unit tests still pass; all 142 server integration tests still pass

Local verification snippet:
```
$ pnpm --filter @moltzap/client test
 Test Files  22 passed | 1 skipped (23)
      Tests  274 passed | 6 todo (280)
$ pnpm --filter @moltzap/server-core test:integration
 Test Files  34 passed (34)
      Tests  142 passed (142)
$ pnpm lint && bash scripts/sloppy-code-guard.sh
Found 0 warnings and 0 errors.
All guards passed.
```

## Deviations from architect plan (justified inline + in commit body)

1. **Conformance file** (`packages/protocol/src/testing/conformance/dispatcher-concurrency.ts`) kept as documentation + `it.todo` placeholders rather than wired into the property registry. Cross-impl wiring requires `TestServer.emitServerRequest`, which would be a public-surface change to a pre-existing module that architect plan §6 ("No public-package-barrel change. No public surface change to pre-existing modules.") explicitly excluded. Properties P1-P4 are exercised at the unit and integration layers against the reference `@moltzap/client` implementation; the file documents the cross-impl contract for a future follow-up that expands `TestServer`.
2. **Integration test (file 5) location**: lives at `packages/client/src/internal/__tests__/s2c-partitioned-dispatcher-real-ws.test.ts` instead of `packages/app-sdk/src/__tests__/dispatcher-concurrency.integration.test.ts` (now removed). The dispatcher behaviour under test is at the `@moltzap/client` layer; `packages/client` already has the in-process WS server helper pattern; `packages/app-sdk` would have needed a new dep on `@effect/platform-node` to bind a server.
3. **`makePartitionedDispatcher`** takes the `Scope.CloseableScope` explicitly as a constructor param rather than deriving it from `Effect.scope`. `Scope.close` requires a `CloseableScope` and `Effect.scope` returns the read-only `Scope.Scope`; deriving the close-handle from the ambient scope would have required either an unsafe cast or a separate `Scope.close` indirection. The caller (`ws-client.ts`) allocates via `Scope.make()` and passes it in. The architect's stub signature (`Effect<…, never, Scope.Scope>`) is preserved for the worker (`makePartitionWorker`); only the dispatcher's signature changed.
4. **Test-update**: one pre-existing `ws-client.test.ts` test was updated to add `sessionId` to its synthetic s2c `params: {}` payload. Pre-#356 the dispatcher passed unparsed params straight to the handler; post-#356 the partition router narrow-validates routing fields (architect §3.2), so synthetic s2c payloads must now include `sessionId` (matching the actual server contract — every `s2cRpcMethods` entry's `params` schema requires it).

## OQ resolutions (architect spec §9)

- **OQ-1** (lifecycle method serialization): default applied — distinct lifecycle methods get distinct `(S, "*lifecycle*", method)` keys → independent fibers. Tenants needing strict cross-method ordering must encode it in their handler. AppHost does not currently require it.
- **OQ-3** (`idlePartitionTtlMs`): 60_000 ms (architect's recommended default).
- **OQ-4** (parked handler exceeds manifest hookTimeout): documented in `s2c-partitioned-dispatcher.ts` header — behaviour matches existing single-fiber dispatcher; the SDK does not interrupt past-deadline handlers; WS-disconnect Scope teardown does.

## Simplify skips
- **none** — the diff was simplified to drop redundant Cause-inspection in worker.offer (single-producer guarantee from reader-fiber makes Queue.size + Queue.offer race-free), drop the `yield_void_exit()` helper in favour of importing `Exit.void` directly, and inline `getOrCreatePartitionWorker`'s scope plumbing into the function body.

## Codex diff review
Pending — codex review was specified as a pre-PR pass but is not blocking landing per the orchestrator's review-357 dispatch plan. Will follow as part of `review-senior-357` + `verify-357`.

## Confidence

**HIGH.** The unit-level deadlock reproducer (`s2c-partitioned-dispatcher.test.ts > "deadlock fix (D10 reproducer at dispatcher level)"`) and the real-WS integration reproducer (`s2c-partitioned-dispatcher-real-ws.test.ts`) both pass under wall-clock conditions where pre-#356 they would hang. The fix is structural (different keys → different fibers), not timeout-based. Effect's `Scope` + `Queue` + forked-fiber primitives map 1:1 to the per-partition lifetime + bounded-buffer + drain-loop requirements; no speculative algorithm choices.


chughtapan (commented):
## VERIFY REQUEST_CHANGES

### CI gate
- ci: SUCCESS
- conformance: SUCCESS
- test-integration: SUCCESS
- Mintlify: SKIPPED

CI green precondition met (HEAD `1b7c3c0`).

### Spec adherence (architect plan #356 r1)
Module/interface/data-flow contract preserved at the macro level: 4 new internal modules under `packages/client/src/internal/` (errors, partition-key, partition-worker, partitioned-dispatcher) + `ws-client.ts` modified, matching plan §2's row table. Public surface change is exactly one optional field on `MoltZapWsClientOptions` (`s2cDispatcher`), as planned.

**One undocumented deviation surfaces below as a P1 finding** — the spec calls for per-worker child scopes (plan §4.2 ASCII: `Scope.close(workerScope) — shuts queue, interrupts fiber`; spec narrative line 364: "the dispatcher Scope's finalization cascades to every worker fiber via their child-scope finalizers"). The implementation reuses one dispatcher scope for every worker, so the reaper cannot run worker finalizers.

### Documented deviations (4)
1. **Conformance file `it.todo`** — defensible. `TestServer.emitServerRequest` is a public surface change to a pre-existing module, excluded by architect §6. P1-P4 properties are exercised against the reference implementation; cross-impl wiring filed as moltzap#358.
2. **Integration test in `packages/client/.../__tests__/`** — defensible. Behaviour under test is at the `@moltzap/client` layer; client already has the in-process WS server pattern; app-sdk would have needed a new `@effect/platform-node` dep.
3. **`makePartitionedDispatcher` takes `Scope.CloseableScope` explicitly** — defensible. `Scope.close` requires a `CloseableScope`; `Effect.scope` returns read-only `Scope.Scope`. Passing the closeable in is the cleanest move.
4. **`ws-client.test.ts` updated to add `sessionId`** — defensible. The partition router narrow-validates routing fields (architect §3.2); synthetic s2c payloads must now include `sessionId` to match the schema contract every `s2cRpcMethods` entry has always required.

### D10 reproducer
Confirmed end-to-end. The unit-level reproducer at `packages/client/src/internal/__tests__/s2c-partitioned-dispatcher.test.ts:313-371@1b7c3c0` parks `apps/onBeforeDispatch` on `Deferred.await(release)` and concurrently offers `apps/onBeforeMessageDelivery` for the same `(SESSION_A, CONV_X)` — the second handler completes, fires the `Deferred`, the first resumes. Pre-#356 vitest's `testTimeout` would catch this as a hang. The real-WS integration test at `s2c-partitioned-dispatcher-real-ws.test.ts:96-218@1b7c3c0` exercises the same scenario through full transport (no mocks). Both pass locally; structural fix (different `hookKind` → different key → different fiber) is correct.

### Codex pass — REQUEST_CHANGES, 5 findings folded inline

**P1 [`packages/client/src/internal/s2c-partitioned-dispatcher.ts:411-423@1b7c3c0`] — Worker finalizers leak on reap; undocumented spec deviation.** Workers are constructed via `Scope.extend(makePartitionWorker(...), state.dispatcherScope)`, so every worker's `Effect.addFinalizer(Queue.shutdown(queue))` (`s2c-partition-worker.ts:170-172@1b7c3c0`) registers directly on the dispatcher scope. The reaper at `s2c-partitioned-dispatcher.ts:494-502@1b7c3c0` only calls `Fiber.interrupt(worker.fiber)`; the queue-shutdown finalizer remains pinned to the dispatcher scope until the connection closes. Architect plan §4.2 ASCII explicitly says `Scope.close(workerScope) — shuts queue, interrupts fiber`, and spec narrative says "cascades to every worker fiber via their child-scope finalizers" — this is a real, undocumented deviation. The consequence is a slow finalizer + Queue-state leak per reaped partition (default 256-partition cap × hours-long connections × ~500B per leaked finalizer pair). Fix: allocate a `Scope.fork(state.dispatcherScope, ...)` per worker, hand it to `makePartitionWorker`, and close that child scope from the reaper instead of interrupting the fiber.

**P1 [`packages/client/src/internal/s2c-partitioned-dispatcher.ts:469-503@1b7c3c0`] — Reaper race silently drops one in-flight request.** The reaper snapshots `idleSince` at lines 477-485, then later removes the key + interrupts the fiber at lines 487-502 without re-checking idleness inside `Ref.update`. A reader-fiber offer that observes the still-mapped worker, sets `idleSince=Busy`, and `Queue.offer`s before the reaper's removal step will succeed at the offer level but the request will never drain (fiber gets interrupted; queue.shutdown only fires at connection close). The author's docstring at lines 460-468 acknowledges the race with "one dropped request retried by the next reader-fiber offer" — but there is no retry; the request sits in the queue and the server-side `Deferred.await` times out at AppHost's `hookTimeout` (5s default). The race window is microseconds within a 60s+5s timeline so prevalence is low, but the documented mitigation is wrong. Combine the fix with the P1 above by re-checking `idleSince._tag === "Idle"` inside the `Ref.update` lambda, or by closing per-worker scopes which surfaces a cleaner "is this still ours" decision.

**P2 [`packages/client/src/internal/s2c-partition-key.ts:52-61@1b7c3c0`] — Schema drift between protocol and partition key extractor.** `CONVERSATION_BEARING_METHODS` and `LIFECYCLE_METHODS` are hardcoded sets disjoint from `s2cRpcMethods` in the protocol. Adding a new s2c method without updating both places compiles cleanly but makes the dispatcher reject every instance as `unknown-method` → `-32602`. Architect §3.2 calls this out as a deliberate cycle-avoidance choice, but a simple counter-measure exists: a unit test that asserts `LIFECYCLE_METHODS ∪ CONVERSATION_BEARING_METHODS = keys(s2cRpcMethods)`. Without it, this is a maintenance trap that won't surface until production traffic fails.

**P2 (false positive) — Soft-cap overshoot under concurrent allocation.** Codex flagged that distinct-key offers could each pass the size check at line 397 and overshoot the cap. False positive given the single-producer reader-fiber guarantee documented at `s2c-partition-worker.ts:19-25@1b7c3c0` — `dispatcher.offer` is only called from the per-connection reader fiber's `Stream.runForEach` loop (`ws-client.ts:1052-1086@1b7c3c0`), so concurrent allocations cannot occur within one dispatcher.

**P2 (false positive) — `PartitionWorker.offer` could suspend under multi-producer.** Same guarantee resolves it. Worth a one-line code comment at the dispatcher header noting that `concurrent reader-fiber offers compose without a router` (line 23) is a misnomer — there is only ever one reader fiber per dispatcher; the actual property is "the route step is wait-free against the worker drain."

### Caller audit
6 direct constructors of `MoltZapWsClient` outside tests:
- `examples/mountains-or-beaches/src/bot.ts:25`
- `packages/client/src/test/index.ts:39`
- `packages/client/src/test-utils/conformance-adapter.ts:101`
- `packages/client/src/service.ts:260`
- `packages/client/src/cli/transport.ts:302`
- `packages/app-sdk/src/app.ts:148`

None pass the new `s2cDispatcher` field; it is optional with `DEFAULT_PARTITIONED_DISPATCHER_CONFIG` defaults applied. No breaking change for existing consumers. Arena's pinned submodule will absorb this without callsite churn.

### P1-P4 craft notes (safer floor)
- **P1 (Types beat tests):** `PartitionKey` branded; `OfferRejected` discriminated union; `IdleSince` and `AllocationOutcome` discriminated. Compliant.
- **P2 (Validate at boundaries):** `extractPartitionKey` does narrow validation of routing fields; AppHost continues to do per-method schema decode. Compliant.
- **P3 (Errors are typed):** All failure modes via `Data.TaggedError`. The single `throw new Error` at `s2c-partition-key.ts:174@1b7c3c0` is correctly scoped to programmer error (caller bypassed branding). Compliant.
- **P4 (Exhaustiveness):** `writeOfferRejection` (`ws-client.ts:909-933@1b7c3c0`) ends in `default: const _exhaustive: never = err`. Compliant.

### Local sanity
- `pnpm --filter @moltzap/client test` → 22 passed | 1 skipped (23) ; **274 passed | 6 todo (280)**
- `pnpm --filter @moltzap/server-core test:integration` → **34 passed (34) ; 142 passed (142)**
- `pnpm lint` → 0 warnings, 0 errors
- `bash scripts/sloppy-code-guard.sh` → All guards passed

### Verdict
**REQUEST_CHANGES** — the D10 fix is structurally correct and the keying invariant holds, but the impl drops a load-bearing piece of the architect plan: workers should live on per-worker child scopes so the reaper can `Scope.close` them rather than orphaning their finalizers on the dispatcher scope (P1#1 + P1#2 above). Both findings are addressable with one structural change. The schema-drift P2 should ship as a small unit test in the same revision. The other two codex findings are false positives under the single-producer guarantee — worth a header-comment clarification rather than a fix.

D10 is fixed; that part can stand. The follow-ups are surgical, not architectural.

Pinned at `1b7c3c04d0272e11d21a2b615f7f20684ccf6801`.

chughtapan (commented):
## VERIFY moltzap PR #357 — APPROVE (merging)

### CI gate at HEAD `61b7263b`
| Check | Status |
|---|---|
| ci | SUCCESS |
| conformance | SUCCESS |
| test-integration | SUCCESS |
| Mintlify Deployment | SKIPPED (expected) |

### Pre-merge safety
- mergeable: `MERGEABLE`
- review-senior: APPROVE-pending (REQUEST_CHANGES on `1b7c3c0` closed at `61b7263b`)
- senior-357-fix: 3 findings closed inline + 2 new architecturally-significant tests + 2 codex review rounds

### Local sanity (HEAD `61b7263b`, /tmp/verify-357)
| Command | Exit | Result |
|---|---|---|
| `pnpm install --frozen-lockfile` | 0 | clean |
| `pnpm --filter @moltzap/protocol build` | 0 | clean (prereq for client tsc) |
| `pnpm --filter @moltzap/client build` | 0 | clean |
| `pnpm --filter @moltzap/server-core build` | 0 | clean |
| `pnpm --filter @moltzap/client test` | 0 | **277 passed / 6 todo / 22 files / 1 skipped** |
| `pnpm --filter @moltzap/server-core test:integration` | 0 | **142 passed / 34 files** |
| `pnpm lint` | 0 | 0 warnings, 0 errors (oxlint, 359 files) |
| `bash scripts/sloppy-code-guard.sh` | 0 | All guards passed |

### D10 reproducer pin
- `s2c-partitioned-dispatcher.test.ts` — **15/15 pass** (structural acceptance gate, includes the unit reproducer)
- `s2c-partitioned-dispatcher-real-ws.test.ts` — **2/2 pass** (real-WS integration acceptance gate)
- `s2c-partition-worker.test.ts` — 9/9 pass
- `s2c-partition-key.test.ts` — **16/16 pass** (includes the new schema-drift guard)

### review-357 findings closure (diff `1b7c3c04..61b7263b -- packages/client/src/internal/`)
- **P1 #1** worker scope-fork pattern — CLOSED. `getOrCreatePartitionWorker` now calls `Scope.fork(state.dispatcherScope, ExecutionStrategy.sequential)` per worker; `reapIdlePartitions` calls `Scope.close(worker.scope, Exit.void)` (no `Fiber.interrupt`); `PartitionWorker` carries `readonly scope: Scope.CloseableScope`. Lost-race teardown also moved to `Scope.close` instead of `Fiber.interrupt`.
- **P1 #2** reaper race with `Retiring` tag — CLOSED. `IdleSince` extended with `{ _tag: "Retiring" }`. Reaper's `Ref.modify(partitionsRef, …)` lambda atomically (a) flips `MutableRef.set(idleSince, Retiring)` and (b) drops the entry from the map. Producer's `worker.offer` uses `MutableRef.updateAndGet(idleSince, current → current._tag === "Retiring" ? current : Busy)` and refuses to enqueue when it observes `Retiring`, surfacing `PartitionQueueFullError`. Idle-tracking migrated `Ref → MutableRef` so the reaper can read it synchronously inside the `Ref.modify` lambda.
- **P2** schema-drift test — CLOSED. New `describe("extractPartitionKey — schema drift guard against \`@moltzap/protocol\`")` block in `s2c-partition-key.test.ts` iterates `s2cRpcMethods` and asserts every wire method is recognised by `extractPartitionKey` (fails the build if a new RPC is added without updating the literal Set).

### Architecturally-significant new tests (architect §10 + Retiring end-to-end)
- `s2c-partitioned-dispatcher.test.ts` — `"reaped workers release queue finalizers from dispatcher scope"` (leak-pin) — present, passing.
- `s2c-partitioned-dispatcher.test.ts` — `"offer racing the reaper either succeeds against a fresh worker or fails with a typed OfferRejected — never silently drops"` (Retiring tag end-to-end) — present, passing.

### Per-criterion check (D10 acceptance from arena#248 + moltzap#356)
- [x] `before_dispatch` suspended on `Deferred.await` does NOT block `before_message_delivery` for the same `(sessionId, conversationId)` — evidence: `s2c-partitioned-dispatcher.test.ts` unit reproducer passes; `s2c-partitioned-dispatcher-real-ws.test.ts` integration variant passes.
- [x] No silent drops under reaper-vs-offer race — evidence: typed `PartitionQueueFullError` on `Retiring`; covered by the new "offer racing the reaper" test.
- [x] No orphan finalizers on dispatcher scope — evidence: per-worker `Scope.fork` + `Scope.close`; covered by the new "reaped workers release queue finalizers" test.
- [x] Schema drift between `s2cRpcMethods` and partition-key router caught at test time — evidence: new schema-drift guard test.

### Verdict
**APPROVE — merging via `gh pr merge --rebase --delete-branch`.**

PR #375, opened 2026-05-03
fix(conversations): findExistingDm filters archived_at IS NULL (closes #372)

### LINKED ISSUE #372: conversations/create: findExistingDm should filter archived_at IS NULL (or closeSession should not archive DMs)
## Symptom

`apps/closeSession` archives every conversation attached to the session
(`packages/server/src/app/app-host.ts:977-981`) — including 1-on-1 DMs
created via `conversations/create({type: "dm"})` and attached via
`apps/attachConversation`. After the session closes, the DM convId remains
in the `conversations` table with `archived_at` set and the participants
unsubscribed.

When a *new* session calls `conversations/create({type: "dm"})` between
the **same two agents**, `ConversationService.findExistingDm`
(`packages/server/src/services/conversation.service.ts:707-735`)
returns the **archived** convId from the previous session — its SQL
query does not include an `archived_at IS NULL` predicate. The early
return at lines 102-108 then skips the participant-resubscribe path at
lines 169-172.

The new session is now holding a convId for an archived conversation
that nobody is subscribed to. Any `messages/send` against it is rejected
with `ConversationArchived`
(`packages/server/src/services/message.service.ts:100-107`), which the
SDK surfaces as `SendError("Failed to send message to conversation
<id>")`.

## Downstream evidence (arena)

This bit moltzap-arena's repeat-mode game gate (Gate B / arena#255 / D15):
arena uses `conversations/create({type: "dm"})` to create per-player role
DMs between the GM and each player, then attaches them to the session.
On the *second* game in `ARENA_LIVE_REPEAT=true` mode with the same agent
roster, every guard correction DM and night-private narration silently
fails because the new session inherits the archived game-1 role DMs.
Day-vote stalls indefinitely waiting for a wolf re-vote that never
arrives. Full repro + log lines on chughtapan/moltzap-arena#255.

## Why it's a moltzap-side defect

Per-session DM uniqueness for the same `(creatorAgentId, otherAgentId)`
pair is the explicit purpose of `findExistingDm` — but DMs typically
outlive any single session, and `closeSession` archiving them flattens
the lifetime invariant. Two reasonable fixes, both architectural choices
that need an architect call:

1. **Filter `findExistingDm` to active DMs.** Add `AND
   c.archived_at IS NULL` to the SQL at lines 715-728. Returns null for
   archived DMs, so the next `conversations/create` falls through to
   the `INSERT INTO conversations …` branch + the `subscribeAgentsToConversation`
   call at lines 169-172. The archived DM stays in the DB for audit /
   message history, but is no longer reused.

2. **Don't archive DMs on `closeSession`.** Treat DMs as long-lived
   conversations that survive the session lifecycle; archive only the
   manifest conversations the session actually owns. This requires a
   `conversation_type`-aware predicate in the `closeSession` archive
   query at app-host.ts:977-981 and is the more invasive option.

Either is fine; (1) is the smaller change. If you keep the dedup, you
also want to think about what to do with archived DMs on dedup hit:
unarchive + resubscribe, or fall through to a new conversation row.

## Arena workaround (already shipped)

moltzap-arena has shipped a 1-line workaround in
`packages/server/src/werewolf-app/hooks/on-session-active.ts`:
request `type: "group"` instead of `type: "dm"` for role DMs.
`findExistingDm` only runs for `type === "dm"`, so groups always create
a fresh conversation. Tracking PR: chughtapan/moltzap-arena#221 (commit
landing for D15 closure).

The arena workaround is cosmetic — role DMs become 2-person groups (GM +
player), same isolation, same lifetime. The proper fix is moltzap-side
and unblocks DM consumers more broadly.

## Suggested modality

`/safer:architect` — schema/lifecycle invariant decision. Either fix
touches dedup semantics or session/conversation lifetime, both of which
need an explicit invariant statement in the architect plan.

## References

- arena#255: https://github.com/chughtapan/moltzap-arena/issues/255
- arena PR #221: https://github.com/chughtapan/moltzap-arena/pull/221
- moltzap server-core paths cited above (HEAD `96bc8104`):
  - `packages/server/src/services/conversation.service.ts:102-108, 169-172, 707-735`
  - `packages/server/src/app/app-host.ts:977-981`
  - `packages/server/src/services/message.service.ts:100-107`

chughtapan:
## Summary

Fixes [arena#372](https://github.com/chughtapan/moltzap/issues/372) — `conversations/create` returned an archived DM and reused it. After `closeSession` archives a DM, the next `conversations/create` for the same pair returned the archived row, but writes to it (`messages/send`) are rejected on archived conversations. The caller ended up with a conversation id it could never write to.

**Fix (Option A):** `findExistingDm` filters `c.archived_at IS NULL` in the SQL lookup. Archived DMs are dead — a fresh conversation is created for the next pairing instead.

This matches the repo's existing convention: `archived_at IS NULL` means live, non-NULL means archived (see the `archive`/`list` paths in the same service).

## Test plan

Added `ConversationService.create — archived DM lookup (issue #372)` in `conversation.service.test.ts`:

- [x] Creates a DM between alice and bob
- [x] Archives it
- [x] Creates a DM again with the same participants
- [x] Asserts the new DM has a different id (not the archived one)
- [x] All 6 tests in `conversation.service.test.ts` pass against PGlite
- [x] `pnpm build` clean
- [x] `pnpm lint` 0 warnings, 0 errors
- [x] `pnpm format` clean
- [x] `scripts/sloppy-code-guard.sh` passes
- [x] Codex review: pass (verified the SQL filter is correct and the test covers the regression)

### DIFF
--- a/packages/server/src/services/conversation.service.test.ts
+++ b/packages/server/src/services/conversation.service.test.ts
@@ -234,3 +234,44 @@ describe("ConversationService.addParticipant auto-subscribes the new member", ()
     expect(bobConn.conversationIds.has(conv.id)).toBe(true);
   });
 });
+
+/**
+ * Regression for arena#372: archived DMs must not be reused by
+ * conversations/create. Before this fix, `findExistingDm` returned any DM
+ * between the two agents — including archived ones. After closeSession
+ * archives a DM, the next conversations/create would hand back the archived
+ * conversation, but messages/send rejects writes to archived conversations,
+ * so the agent ended up with a "live" DM id it could never write to.
+ */
+describe("ConversationService.create — archived DM lookup (issue #372)", () => {
+  beforeEach(freshDb, dbHookTimeoutMs);
+  afterEach(async () => {
+    await pglite?.close();
+  });
+
+  it("does not reuse an archived DM — creates a fresh conversation instead", async () => {
+    const connections = new ConnectionManager();
+    const participants = new ParticipantService(db);
+    const service = new ConversationService(db, participants, connections);
+    const authService = new AuthService(db);
+
+    const alice = await seedAgent(authService, "alice");
+    const bob = await seedAgent(authService, "bob");
+
+    // 1. Create a DM between alice and bob.
+    const first = await Effect.runPromise(
+      service.create("dm", undefined, [bob], alice),
+    );
+
+    // 2. Archive it (alice is the creator/owner, so archive is permitted).
+    await Effect.runPromise(service.archive(first.id, alice));
+
+    // 3. Create a DM again with the same participants.
+    const second = await Effect.runPromise(
+      service.create("dm", undefined, [bob], alice),
+    );
+
+    // 4. The new DM must be a fresh row — not the archived one.
+    expect(second.id).not.toBe(first.id);
+  });
+});
--- a/packages/server/src/services/conversation.service.test.ts
+++ b/packages/server/src/services/conversation.service.test.ts
@@ -274,4 +274,27 @@ describe("ConversationService.create — archived DM lookup (issue #372)", () =>
     // 4. The new DM must be a fresh row — not the archived one.
     expect(second.id).not.toBe(first.id);
   });
+
+  it("still dedupes live DMs — repeat create returns the existing live DM", async () => {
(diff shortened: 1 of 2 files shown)


chughtapan:
verify-375: SHIP

Issue #370, opened 2026-05-02
protocol: register `uuid` TypeBox format from the package itself so consumers don't have to


chughtapan:
Discovered while fixing arena#253 (commit `1f162c1` on `staff/217-arena-rewrite-app-sdk-client`).

`PresenceChangedEventSchema` and other protocol schemas use `Type.String({ format: "uuid" })` (via `brandedId()` in `packages/protocol/src/helpers.ts`), but no code in moltzap's `protocol`, `app-sdk`, `client`, or `server-core` packages calls `FormatRegistry.Set("uuid", ...)`. Consumers (like arena-server) must register the format themselves before any `Value.Decode` runs, or every frame is rejected with `Unable to decode value as it does not match the expected schema; inner: { type: Object, message: 'Unknown format uuid', path: '/agentId' }`.

Concrete impact: arena's agent-presence watcher's `onlineSet` stayed empty across 10 verify cycles because every `presence/changed` frame failed schema decode. The arena-side workaround is a side-effect import at the bin entrypoint:

```ts
import { FormatRegistry } from "@sinclair/typebox";
const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
if (!FormatRegistry.Has("uuid")) {
  FormatRegistry.Set("uuid", (v) => typeof v === "string" && UUID_RE.test(v));
}
```

That works for the bin path but is fragile — any consumer that imports `arena-harness` or `create-app` directly without going through `index.ts` (e.g., the eval harness path `packages/evals/src/start-agents.ts` → `arena-harness.ts` dynamic-imports `create-app.js`) bypasses the registration and re-hits the bug.

Suggested fix: add `packages/protocol/src/register-formats.ts` that registers `uuid` (and any other formats the protocol uses, e.g., `date-time`, `uri` if applicable) as a side-effect at module load, and re-export it from the package index so `import "@moltzap/protocol"` registers automatically. That makes every consumer correct by construction.

Modality recommendation: `architect` (this is a protocol-package decision; touches the public surface and load-order semantics).

Cross-refs:
- arena#253 — root-cause writeup
- moltzap-arena PR #221 — server-local workaround landed in commit `1f162c1`

chughtapan:
## Audit verdict: DONE

**Audited:** 2026-05-08T04:00Z against `main` at `a964113`
[...]

Issue #216, opened 2026-04-24
messages list prints literal 'undefined' in seq column


chughtapan:
Discovered during `/safer:verify` of sbd#192 (CLI v2 DoD E2E dogfood) at moltzap main `2a523d8`.

## Symptom

```
$ moltzap --as $ALICE_KEY messages list --conversation <convId>
undefined	fe9a622b-…	hello from bob via ws
```

## Root cause

`packages/client/src/cli/commands/messages.ts` prints `\`\${m.seq}\\t\${sender}\\t\${text}\`` but `MessageSchema` in `packages/protocol/src/schema/messages.ts` has no `seq` field. Server response is correct per schema; the CLI reads an undefined field and stringifies it.

## Fix options

- Remove the seq column from the print format (cosmetic fix).
- OR add `seq` to `MessageSchema` and the server-side `mapMessage` in `message.service.ts` (spec change — orchestrate via `/safer:spec`).

## Severity

Cosmetic. All 14 CLI v2 DoD items otherwise ship-clean; flagged in the verify verdict as SHIP_WITH_CONCERNS.

## Refs

- Verify verdict: https://github.com/chughtapan/safer-by-default/issues/192#issuecomment-4313824827
- Related: #214 (messages list cursor support)

chughtapan:
## Audit verdict: DONE

**Audited:** 2026-05-08T04:00Z against `main` at `a964113`
**Auditor:** orchestrate-dispatched repo-wide cleanup audit (epic #512 Amendment 6)

### Symptom check

- [x] CLI no longer prints literal `undefined` for `seq` — `packages/client/src/cli/commands/messages.ts:54-55,78@a964113` print loop is `${m.createdAt}\t${sender}\t${text}` (no seq column); top-of-file comment names PR #379 as the removal site (`messages.ts:54-58@a964113`).
- [x] `seq` removed from wire shape — protocol no longer exposes `seq` on the `MessagesList` result; the field survives only as a server-internal monotonic ordering token in `packages/server/src/services/message.service.ts:214,239,440-442@a964113` (used for cursor/since-seq, not transmitted as a column the CLI reads).

### Disposition

PR #379 ("chore: small mechanical fixes (closes #352 #216 #370)") merged 2026-05-03; the natural-language closure list did not auto-close this issue but the work shipped. The CLI's print path matches the issue's "Remove the seq column from the print format (cosmetic fix)" option. Verified at head.

### Closing as: completed

Issue #213, opened 2026-04-24
examples/mountains-or-beaches: pre-existing tsc --noEmit failure (missing @moltzap/app-sdk, implicit-any)


chughtapan:
Surfaced by staff-187 during sbd#187 cc-channel impl work on PR #209. Pre-existing on `arch/claude-code-channel` branch before the staff commit; required `--no-verify` to land.

## Symptoms

`tsc --noEmit` fails in `examples/mountains-or-beaches` with:
- missing `@moltzap/app-sdk` (deps?)
- implicit-any errors

## Fix direction

Either (a) add missing deps / type annotations to get examples/mountains-or-beaches clean, or (b) exclude examples/ from the pre-commit tsc scan.

## Context

- Blocks clean pre-commit path for contributors touching the monorepo
- Same class as chughtapan/moltzap#208 (different module)
- Consuming PR: https://github.com/chughtapan/moltzap/pull/209

chughtapan:
## Audit verdict: OBSOLETED

**Audited:** 2026-05-08T04:00Z against `main` at `a964113`
**Auditor:** orchestrate-dispatched repo-wide cleanup audit (epic #512 Amendment 6)

### Symptom check

- [x] `examples/mountains-or-beaches` no longer exists — the entire `examples/` directory has been removed from main (`ls examples/` returns "No such file or directory" at `a964113`). The `tsc --noEmit` failure surface this issue named is moot because the file tree it pinned is gone.

### Disposition

Sibling repo cleanup superseded the issue: `examples/mountains-or-beaches` was deleted rather than fixed. There is no current consumer to type-check, so neither fix path (add deps / add types, or exclude examples/ from pre-commit) applies. Pre-commit currently runs clean on main.

### Closing as: not planned

Issue #208, opened 2026-04-24
packages/runtimes/src/trace-capture-harness.ts: pre-commit hook fails on implicit-any errors on clean main


chughtapan:
Surfaced by sbd#185 architect work on the CLI v2 stubs PR (#207). On a clean origin/main checkout, the pre-commit hook fails with implicit-any errors in `packages/runtimes/src/trace-capture-harness.ts`. Blocks any contributor from pushing cleanly without `core.hooksPath=/dev/null` bypass.

## Fix direction

Either (a) add explicit type annotations to the implicit-any sites in trace-capture-harness.ts, or (b) tighten the tsconfig that governs pre-commit for this package.

## Context

- Full architect escalation: https://github.com/chughtapan/safer-by-default/issues/177#issuecomment-4311825849 (Q-PC-1)
- Consuming sub-issue: https://github.com/chughtapan/safer-by-default/issues/185

chughtapan:
Promoting from KEEP-OPEN to DISPATCH-ready (junior-tier). This blocks contributors from pushing cleanly without `core.hooksPath=/dev/null` bypass — it's not cleanup, it's an active pre-commit gate failure on main. Fix recipe is in the body: either add explicit type annotations to implicit-any sites in `packages/runtimes/src/trace-capture-harness.ts` OR tighten the package's pre-commit tsconfig. ~30m. Same class as #213 (`examples/mountains-or-beaches` tsc fail).

chughtapan:
## Audit verdict: DONE

**Audited:** 2026-05-08T04:00Z against `main` at `a964113`
**Auditor:** orchestrate-dispatched repo-wide cleanup audit (epic #512 Amendment 6)

### Symptom check

- [x] `packages/runtimes/src/trace-capture-harness.ts` no longer has implicit-any sites — file is fully type-annotated at `packages/runtimes/src/trace-capture-harness.ts:1-200@a964113` (every parameter, return, and structural shape declared via interfaces; `Effect.Effect<T, E, R>` channels typed throughout).
- [x] Pre-commit hook passes on clean main — `.husky/pre-commit@a964113` runs `pnpm check && pnpm build && pnpm typecheck && ./scripts/sloppy-code-guard.sh`; recent merged PRs (#510, #515) landed without `--no-verify`, evidence the gate is green.
[...]

PR #379, opened 2026-05-03
chore: small mechanical fixes (closes #352 #216 #370)

### LINKED ISSUE #352: [safer:implement-junior] withPatchedProcessEnv loses keys when processEnv === process.env (compose YAML interpolation broken)
**Source:** verify-221-v2 4-player game gate on chughtapan/moltzap-arena PR #221 (arena's E.1 rewrite). Verify's full evidence at https://github.com/chughtapan/moltzap-arena/pull/221#issuecomment-4358674880

## Bug

`lib/moltzap/packages/server/src/runtime-surface/config.ts`'s `withPatchedProcessEnv(processEnv, …)` is called with `processEnv === process.env`. `replaceProcessEnv()` then deletes every key from `process.env` and iterates `Object.entries(next)` over the now-empty same reference, leaving `process.env` empty for the duration of the YAML load.

The interpolator in `lib/moltzap/packages/server/src/config/loader.ts:51` reads `process.env[varName]` and reports the variable missing — even when the var IS in the container's process env at boot.

## Symptom

`moltzap.yaml` with `${REGISTRATION_SECRET}` (or any other `${ENV_VAR}` interpolation) fails with `RuntimeConfigSurfaceError → EnvironmentInvalid <VAR>` even when the env var is set in the container.

verify-221-v2 hit this on docker-compose boot:
- `compose/moltzap.yaml` has `registration.secret: ${REGISTRATION_SECRET}`
- Container has `REGISTRATION_SECRET=<value>` set
- moltzap-network exits with `EnvironmentInvalid REGISTRATION_SECRET`

Verifier worked around by hard-coding the secret in `compose/moltzap.yaml` to unblock the rest of the stack — not acceptable for ship.

## Fix

`withPatchedProcessEnv` should NOT self-alias. Either:
- Snapshot `process.env` to a separate object BEFORE deleting keys.
- OR don't iterate the same reference being mutated — copy first, then delete-and-replace.

Pseudocode:
```ts
function replaceProcessEnv(next: NodeJS.ProcessEnv): void {
  const snapshot = { ...next };  // copy BEFORE mutation
  for (const key of Object.keys(process.env)) delete process.env[key];
  for (const [k, v] of Object.entries(snapshot)) process.env[k] = v;
}
```

Add a unit test that calls `withPatchedProcessEnv(process.env, ...)` (same reference) and confirms `process.env` is unchanged afterward.

## Acceptance

1. `lib/moltzap/packages/server/src/runtime-surface/config.ts` — `withPatchedProcessEnv` doesn't lose keys when `processEnv === process.env`.
2. Unit test pins the regression.
3. arena's compose stack runs `compose/moltzap.yaml` with `${REGISTRATION_SECRET}` interpolation working.

## Modality

`safer:implement-junior` — single-module fix; localized; ~10 LOC.

## Parent
- arena PR #221 verify-221-v2 evidence: https://github.com/chughtapan/moltzap-arena/pull/221#issuecomment-4358674880
- arena parent epic: https://github.com/chughtapan/moltzap-arena/issues/198

## Status
`planning`

chughtapan:
Cluster of three small mechanical fixes — one PR per the assignment.

## Issues closed

- **#352** — `withPatchedProcessEnv` loses keys when `processEnv === process.env` (compose YAML interpolation broken)
- **#216** — `messages list` prints literal `undefined` in the seq column
- **#370** — `@moltzap/protocol`: register `uuid` TypeBox format from the package itself so consumers don't have to

## Issues NOT touched (already resolved on main)

- **#208** — `trace-capture-harness.ts` implicit-any: file already has explicit type annotations everywhere; `pnpm typecheck` is clean.
- **#213** — `examples/mountains-or-beaches` tsc failure: `pnpm typecheck` excludes `./examples/**` already (option (b) from the issue's fix-direction); `pnpm build` builds the example cleanly when run after deps.

### #352 — `packages/server/src/runtime-surface/config.ts`
`replaceProcessEnv(next)` snapshots `next` with `{ ...next }` before deleting keys from `process.env`. Previously, when `next === process.env`, deleting all keys from `process.env` also emptied `next` (same reference), so the re-population loop ran over an empty object and `process.env` stayed wiped for the duration of the YAML load. The interpolator then saw an empty `process.env` and reported every `${VAR}` as missing — exactly the failure verify-221-v2 hit on docker-compose boot.

Regression test: `packages/server/src/runtime-surface/config.test.ts` writes a YAML config with `${VAR}` interpolation, calls `loadRuntimeProcessConfig({ processEnv: process.env })`, and asserts the var is interpolated and `process.env` survives. The test fails on the un-fixed code (verified by stashing the fix).

### #216 — `packages/client/src/cli/commands/messages.ts`
`MessageSchema` (`packages/protocol/src/schema/messages.ts`) has no `seq` field. The CLI was printing `${m.seq}\t${sender}\t${text}`, stringifying `undefined` into the leading column. Switched the leading column to `m.createdAt` (which is in the schema and gives the same ordering signal) and dropped `seq` from the internal `WireMessage` type. Updated the existing unit test to assert `firstLine.startsWith("undefined\\t")` is `false`.

### #370 — `packages/protocol/src/helpers.ts`
Register `uuid`, `date-time`, and `uri` formats with TypeBox's `FormatRegistry` as a side effect at module load, gated on `FormatRegistry.Has(...)` so a consumer-provided stricter validator (registered before this module loads) is preserved. `date-time` pairs the shape regex with `Date.parse` so values like `"2026-99-99T99:99:99Z"` reject. Regression tests in `helpers.test.ts` exercise each format via `Value.Check`.

## Codex review pass

Ran `codex exec` on the diff. Findings addressed:
1. ✅ `date-time` shape-only validation — paired with `Date.parse`.
2. ✅ `FormatRegistry.Set` clobbers consumer registrations — gated on `Has(...)`.
3. ✅ `uri` test only asserted `Has` — added accept/reject cases.
4. ✅ `config.test.ts` blindly deleted env keys in afterEach — now saves and restores prior values.

One pre-existing concern flagged by codex (test fixture in `messages.test.ts` doesn't strictly satisfy `MessageSchema` — missing `conversationId`, includes non-schema `senderName`) — pre-existing, separate from this fix, deferred.

## Verification

- `pnpm build` — clean
- `pnpm typecheck` — clean
- `pnpm format:check` — clean
- `pnpm test` — all passing
- `./scripts/sloppy-code-guard.sh` — all guards passed

### DIFF
--- a/packages/client/src/cli/commands/messages.test.ts
+++ b/packages/client/src/cli/commands/messages.test.ts
@@ -57,7 +57,6 @@ describe("messages list", () => {
       messages: [
         {
           id: "m1",
-          seq: 1,
           senderId: "a1",
           senderName: "alice",
           createdAt: "2026-04-24T00:00:00Z",
@@ -65,7 +64,6 @@ describe("messages list", () => {
         },
         {
           id: "m2",
-          seq: 2,
           senderId: "b1",
           senderName: "bob",
           createdAt: "2026-04-24T00:00:01Z",
@@ -84,6 +82,12 @@ describe("messages list", () => {
       params: { conversationId: "c1", limit: 50 },
     });
     expect(stdout).toHaveBeenCalledTimes(2);
+    // Regression #216: first column is `createdAt`, never `undefined`.
+    // MessageSchema has no `seq` field; the previous output stringified
+    // `m.seq` as the literal "undefined" in the leading column.
+    const firstLine = String(stdout.mock.calls[0]?.[0] ?? "");
+    expect(firstLine.startsWith("undefined\t")).toBe(false);
+    expect(firstLine).toBe("2026-04-24T00:00:00Z\talice\thello");
   });
 
   it("omits limit when absent", async () => {
--- a/packages/client/src/cli/commands/messages.ts
+++ b/packages/client/src/cli/commands/messages.ts
@@ -48,7 +48,6 @@ export interface MessagesListArgs {
 
 interface WireMessage {
   readonly id: string;
-  readonly seq: number;
   readonly senderId: string;
   readonly senderName?: string;
   readonly createdAt: string;
@@ -78,7 +77,7 @@ export const messagesListHandler = (
       for (const m of result.messages) {
         const text = m.parts.find((p) => p.type === "text")?.text ?? "";
         const sender = m.senderName ?? m.senderId;
-        console.log(`${m.seq}\t${sender}\t${text}`);
+        console.log(`${m.createdAt}\t${sender}\t${text}`);
       }
       if (result.hasMore) {
         console.log("... more messages available");
--- a/packages/protocol/src/helpers.test.ts
+++ b/packages/protocol/src/helpers.test.ts
@@ -1,6 +1,8 @@
 import { describe, expect, it } from "vitest";
 import Ajv from "ajv";
 import addFormats from "ajv-formats";
+import { FormatRegistry, Type } from "@sinclair/typebox";
+import { Value } from "@sinclair/typebox/value";
 import { stringEnum, brandedId, DateTimeString } from "./helpers.js";
 
 const ajv = addFormats(new Ajv({ strict: true }));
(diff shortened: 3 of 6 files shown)


chughtapan:
## verify ✅ SHIP

**Gates:** ci ✅ conformance ✅ test-integration ✅ · sloppy-guard 0 · local pkg tests pass

**Acceptance:**
- **#352** withPatchedProcessEnv — snapshot-before-mutate verified; regression test covers `processEnv === process.env` and the separate-snapshot path
- **#216** messages "undefined seq" — `WireMessage` is a CLI-local interface, not the wire schema; `MessageSchema` in protocol never had `seq`. Removing the phantom field is **not** a public surface change. Test pins `firstLine.startsWith("undefined\t")` regression
- **#370** TypeBox uuid/date-time/uri — registered via `FormatRegistry.Has()` guard; `Has()` is the correct API for TypeBox 0.34.49 (independently confirmed by codex). date-time runs `Date.parse` so semantically-impossible-but-shape-valid strings reject. Tests cover all three formats + the semantic check
- **#208** (trace-capture-harness) and **#213** (mountains-or-beaches) — verified already-resolved on main: `tsc --noEmit` clean once workspace deps are built first

**Independent codex review** (high reasoning, no prior context): 0 BLOCKER/HIGH. Two NITs (≤OQ1, not ship-blocking):
- `messages.ts:60` TSDoc still mentions "seq, senderName, text" — actual emit is `createdAt`
- Has()-guard "do not clobber" behavior not directly pinned by test (guard correctness verified externally)

Stamina ≥3 (junior + codex-in-PR + verify-codex). Merging.

PR #378, opened 2026-05-03
fix(server-core): direct conversations honor contact policy (closes #361)

chughtapan:
## Summary

`ConversationService.create("dm", ...)` (and the `messages/send { to: "agent:<name>" }` path through `createDmByAgentName`) was inserting DM participants without consulting the contact-policy `ContactService` that AppHost already uses for app-session admission. Two agents whose owners had no allowed-contact edge could open a DM and exchange messages — bypassing the experiment's communication topology.

Closes #361.

## What changed

- **`AppHost.getContactService()`** (new) exposes the post-construction `setContactService(...)` value to peer services so the policy is defined in one place.
- **`ConversationServiceLive`** wires a *lazy* resolver into the service — lazy because `app.setContactService(...)` runs in `standalone.ts` AFTER the layer has already produced its instance, so a captured snapshot would always be `null`.
- **`ConversationService.create`** now reads `owner_user_id` for each participant in the existing existence-check query (no extra round-trip), and for `type === "dm"` calls the policy after `findExistingDm` (so previously-created DMs stay accessible — idempotent semantics preserved).
- **Typed errors** (Principle 3): failures use `RpcFailure` with the dedicated `ErrorCodes.NotInContacts` code via the new `notInContacts(...)` factory in `runtime/errors.ts`. Missing `owner_user_id` on either side fails closed with the same code, matching `AppHost.checkIdentity`'s stance for app-session admission.

## Behavior matrix

| Scenario                                          | Before        | After                        |
| ------------------------------------------------- | ------------- | ---------------------------- |
| No policy configured (dev/local)                  | DM created    | DM created (unchanged)       |
| Policy set, owners NOT in contact                 | **DM created (bug)** | **`NotInContacts` denial**   |
| Policy set, owners in contact                     | DM created    | DM created                   |
| Policy set, either agent owner-less               | DM created    | `NotInContacts` denial       |
| Existing DM (idempotent re-create)                | Returned      | Returned (policy not re-run) |

## Tests

New PGlite-backed unit tests in `conversation.service.test.ts` lock the contract at the service boundary (per the existing test docstring's pattern):

- DENY when policy returns `false` — asserts `NotInContacts` code + policy invocation args
- SUCCESS when policy returns `true` — DM created, sockets subscribed
- DENY when either agent is owner-less
- ALLOW when no policy is configured (preserves dev-mode default)
- IDEMPOTENT — second create returns the existing DM without re-invoking the policy

## Scope notes

[...]

Issue #361, opened 2026-05-02
[safer:spec] bug(server-core): direct conversations must honor contact policy


chughtapan:
Parent: https://github.com/chughtapan/moltzap/issues/360
Modality: spec
Depends on: none
Acceptance: A spec comment defines the communication policy for direct DMs, group creation, group participant adds, and `messages/send` auto-DM; it names every RPC path to gate, the identity/contact model, expected error behavior, and conformance coverage.

## Context

The experiment requires a contact service to ensure that only explicitly provided pairs of agents can communicate directly. Today, direct conversation creation appears to bypass `ContactService`: `ConversationService.create` validates agent existence and DM/group shape, then inserts participants, but does not call contact policy. `messages/send` can also auto-create a DM from `to: agent:<name>`.

Relevant code references from the orchestration pass:

- `packages/server/src/services/conversation.service.ts` — `ConversationService.create` creates DM/group participants without contact checks.
- `packages/server/src/app/handlers/messages.handlers.ts` — `messages/send` resolves `to: agent:<name>` through `createDmByAgentName`.
- `packages/server/src/app/app-host.ts` — `ContactService.areInContact` is currently checked only during AppHost admission.
- Existing related contacts conformance tracker: <https://github.com/chughtapan/moltzap/issues/335>

## Required Spec Questions

- Should contact policy be user-owner based, agent-pair based, or both?
- Which direct communication paths must be gated: `conversations/create`, `conversations/add-participant`, `messages/send { to }`, existing-conversation sends, app-session conversations, or all of the above?
- What is the correct failure code and user-facing error for a disallowed edge?
- How should dev/local mode represent distinct synthetic owners so topology edges can be tested?
- What conformance cases prove that non-edges cannot communicate while allowed edges can?

## Status

`planning`

PR #387, opened 2026-05-03
fix(server-core): contact policy gates groups + addParticipant (#361 follow-up)

chughtapan:
## Summary

Follow-up to #378 (which gated DM creation on contact policy): extends the same gate to **group creation** and **`conversations/addParticipant`**. Without these, an admin/owner could side-channel an out-of-contact agent into the topology by either:

1. Creating a `type: "group"` conversation with that agent as a member at create time, or
2. Calling `addParticipant` to drag them into an existing group.

This was flagged by the team lead's "be thorough" directive on #361 — fix obvious bypass paths in scope while in the area.

Refs #361.

## Behavior matrix (pairwise model)

Every (requester, target) edge added to a conversation must be allowed by the policy. Group members aren't transitively trusted — the creator/requester is the contact-policy peer for every edge. Owner-less agents fail closed with `NotInContacts`, matching `AppHost.checkIdentity`.

| Path                                | Before #378 | After #378 (DM only) | After this PR    |
| ----------------------------------- | ----------- | -------------------- | ---------------- |
| `create({ type: "dm" })`            | bypass      | gated                | gated (#378)     |
| `messages/send { to }` auto-DM      | bypass      | gated (via #378)     | gated (#378)     |
| `create({ type: "group" })`         | bypass      | **bypass**           | **gated**        |
| `addParticipant(...)`               | bypass      | **bypass**           | **gated**        |

## Implementation

- `requireDirectAllowed` → **`requireCreatorContactsAll`**: loops over each member id and runs the policy on the (creator, member) edge. DM is the one-element case; group is the N-element case.
- New **`checkContactEdge`** helper: shared by `create` and `addParticipant`. Single place that handles owner-null fail-closed + policy invocation + denial logging.
- **`addParticipant`** resolves the requester's owner_user_id once (via `participants.resolve`), reuses the target owner_user_id already returned by `requireExists`, and runs `checkContactEdge` before the participant insert.
- Tightened the owner-id Map invariant in `requireCreatorContactsAll` — explicit `has()` guard with a `NotFound` failure surfaces an upstream invariant break instead of silently downgrading to a denial via `?? null`.

## Tests

10 new PGlite-backed tests (20 in the suite total):

- **Group**: deny when ANY (creator, member) edge fails; allow when all edges pass; fail-closed when any member is owner-less. Policy-call array assertion proves the loop runs both edges before the deny (no accidental short-circuit).
- **`addParticipant`**: deny when (requester, target) denied; allow when policy permits; fail-closed when either is owner-less; default unrestricted when no policy.
- **`createDmByAgentName`**: pinned that the auto-DM path still triggers the gate (per senior review on #378).

## Reviews completed (stamina N≥3)

Three independent reviews, full results in PR #378 conversation:

- **Senior review** (independent agent, no prior context): SHIP-WITH-MINOR-FIXES, both fixes addressed.
- **Codex review** (`codex exec -c model_reasoning_effort=high`): SHIP, no blocking findings.
- **Architectural review** (cold-eyes): NEEDS-EXTRACTION-LATER — ship now, but flagged the layering smell + duplication for a follow-up extraction before the next gate (`apps/attachConversation`). Filed as known follow-up — not blocking this PR.

## Test plan

- [x] `pnpm --filter @moltzap/server-core test src/services/conversation.service.test.ts` — 20/20 passing
- [x] `pnpm --filter @moltzap/server-core build` — clean
- [x] `pnpm lint` — 0/0
- [x] `pnpm format` — clean
- [x] `scripts/sloppy-code-guard.sh` — all guards passed
- [x] Pre-commit hooks (lint + typecheck) pass

### DIFF
--- a/packages/server/src/services/conversation.service.test.ts
+++ b/packages/server/src/services/conversation.service.test.ts
@@ -508,3 +508,300 @@ describe("ConversationService.create enforces contact policy on DMs", () => {
     expect(calls.length).toBe(1);
   });
 });
+
+/**
+ * Group-creation gate (#361 cleanup pass): a creator cannot drop an
+ * out-of-contact agent into a group conversation. The policy runs
+ * pairwise — every (creator, member) edge must be allowed. Group members
+ * are NOT transitively trusted; the creator is the requester for every
+ * edge.
+ */
+describe("ConversationService.create enforces contact policy on groups", () => {
+  beforeEach(freshDb, dbHookTimeoutMs);
+  afterEach(async () => {
+    await pglite?.close();
+  });
+
+  it("denies group creation when ANY (creator, member) edge fails", async () => {
+    const calls: Array<[string, string]> = [];
+    // Allow alice⇄bob, deny alice⇄carol.
+    const policy = (a: string, b: string) =>
+      Effect.sync(() => {
+        calls.push([a, b]);
+        return b !== "00000000-0000-0000-0000-0000000000c3";
+      });
+
+    const connections = new ConnectionManager();
+    const participants = new ParticipantService(db);
+    const service = new ConversationService(
+      db,
+      participants,
+      connections,
+      undefined,
+      () => policy,
+    );
+    const authService = new AuthService(db);
+
+    const alice = await seedAgent(
+      authService,
+      "alice",
+      "00000000-0000-0000-0000-0000000000a1",
+    );
+    const bob = await seedAgent(
+      authService,
+      "bob",
+      "00000000-0000-0000-0000-0000000000b2",
+    );
+    const carol = await seedAgent(
+      authService,
+      "carol",
+      "00000000-0000-0000-0000-0000000000c3",
+    );
+
+    const exit = await Effect.runPromiseExit(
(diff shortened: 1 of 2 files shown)


chughtapan:
verify-387 (round 2): **SHIP**

Bypass closed by 05cd63a. Re-verified:
- ✅ `addParticipant` rejects `type='dm'` with `InvalidParams` BEFORE policy/insert (`conversation.service.ts:579-592`), placed after `requireRole` so non-participants can't probe DM existence (auth-first).
- ✅ New regression test pins (a) `InvalidParams` rejection and (b) `conversation_participants` table unchanged — Carol never inserted.
- ✅ Symmetric with `leave()`'s existing DM guard.
- ✅ ci / conformance / test-integration all pass.
- ✅ sloppy-guard: 0 violations.
- ✅ codex re-review: SHIP verdict, no blocking findings.

Stamina: senior + senior-cold-eyes + codex-1 + architect-cold-eyes + verify-codex (round 1, HOLD) + verify-codex (round 2, SHIP) = N=6 heterogeneous reviews. Public-surface verified. Merging.

PR #389, opened 2026-05-03
test(client): align messages.test.ts fixture with MessageSchema (deferred from #379)

chughtapan:
## Summary

Five-line fixture fix deferred from PR #379. Codex flagged in that
PR's review that the `messages/list` fake transport fixture in
`packages/client/src/cli/commands/messages.test.ts` omits
`conversationId` — a required field on `MessageSchema` (which has
`additionalProperties: false`). The fix was deferred as out-of-scope
for the original PR.

This PR adds `conversationId` to both messages in the fixture and
refreshes the comment to honestly describe what's there:

- Every required `MessageSchema` field is present (now including
  `conversationId`).
- `senderName` is the CLI display fallback the handler reads
  (`m.senderName ?? m.senderId`); it is **not** part of
  `MessageSchema` itself. The comment now says so explicitly,
  instead of the previous (inaccurate) "server-computed" framing.

## Pre-existing structural concern (NOT addressed — out of scope)

The `WireMessage` type in `packages/client/src/cli/commands/messages.ts`
declares `senderName?: string`, but the actual `messages/list` result
schema (`Type.Array(MessageSchema)`) has `additionalProperties: false`,
so a real server response carrying `senderName` would fail schema
validation. That's a structural client/server mismatch, not a fixture
bug, and is left for a separate change.

## Verification

- `pnpm build` — clean
- `pnpm typecheck` — clean
- `pnpm format:check` — clean
- `pnpm test` (client package) — 277 passed (messages.test.ts: 3 passed)
- `./scripts/sloppy-code-guard.sh` — all guards passed
- `codex exec -c model_reasoning_effort=high` — accepted the
  field placement; comment was tightened in response to its feedback

## Test plan

- [ ] CI green
- [ ] `pnpm --filter @moltzap/client test -- src/cli/commands/messages.test.ts` shows 3 passed

### DIFF
--- a/packages/client/src/cli/commands/messages.test.ts
+++ b/packages/client/src/cli/commands/messages.test.ts
@@ -53,10 +53,15 @@ describe("messages list", () => {
   afterEach(() => stdout.mockRestore());
 
   it("calls messages/list with { conversationId, limit? } and emits one line per message", async () => {
+    // Fixture matches the `messages/list` result shape: every required
+    // `MessageSchema` field is present (including `conversationId`).
+    // `senderName` is the CLI display fallback the handler reads; it is
+    // not part of `MessageSchema` itself (see WireMessage in messages.ts).
     const { calls, transport } = makeFakeTransport(() => ({
       messages: [
         {
           id: "m1",
+          conversationId: "c1",
           senderId: "a1",
           senderName: "alice",
           createdAt: "2026-04-24T00:00:00Z",
@@ -64,6 +69,7 @@ describe("messages list", () => {
         },
         {
           id: "m2",
+          conversationId: "c1",
           senderId: "b1",
           senderName: "bob",
           createdAt: "2026-04-24T00:00:01Z",


chughtapan:
verify-389: SHIP

- CI: ci ✅ conformance ✅ test-integration ✅ (Mintlify skipping per spec)
- Diff: 6-line addition matches scope — `conversationId` added to both fixtures, comment about `senderName` being a CLI display fallback verified against WireMessage in messages.ts
- MessageSchema: confirmed `conversationId` is required (packages/protocol/src/schema/messages.ts:42)
- sloppy-guard: 0 violations
- Codex (1 round): flagged that `"c1"` isn't a valid UUID. Out-of-scope — pre-existing fixture IDs (`m1`, `a1`, `b1`) also non-UUID and untouched; handler uses fake transport with no runtime schema validation; PR's narrow scope (add the missing required field) is achieved.

Merging.

PR #490, opened 2026-05-06
fix(presence): replace-semantics on presence/subscribe (#487)

### LINKED ISSUE #487: presence/subscribe accumulates subscriptions — no unsubscribe path leaks fan-out for long-running clients
Surfaced from a downstream consumer (`moltzap-arena` BYOA migration) during `/plan-eng-review` on 2026-05-05.

## What

`PresenceService.subscribe(connId, agentIds)` at `packages/server/src/services/presence.service.ts:51-58` is **append-only**:

```typescript
subscribe(connId: string, agentIds: ReadonlyArray<string>): void {
  for (const agentId of agentIds) {
    let subs = this.subscribers.get(agentId);
    if (!subs) { subs = new Set(); this.subscribers.set(agentId, subs); }
    subs.add(connId);
  }
}
```

There is no `unsubscribe(connId, agentIds)` method exposed via RPC. The only purge is `removeConnection(connId)` at `presence.service.ts:69-73`, which fires on socket disconnect and removes connId from **every** agent's subscriber set (full purge). Per-call un-subscription is impossible.

## Why this matters

For clients that hold a long-running connection and **periodically re-evaluate which agents to watch** — the canonical pattern for any "live address book" use case — the subscriber set monotonically grows. Each `presence/subscribe([currentSet])` call accumulates `connId` into the union of all `currentSet`s ever subscribed.

Concretely from the arena's BYOA watcher: post-#481 the watcher reads contact-scoped `agents/list` each iteration and re-subscribes to current online-eligible agents. With a fixed-fleet (5 agents, never changes) the union is bounded — the existing watcher works fine. With BYOA (set grows when contacts get accepted, shrinks when they're removed/expired) the union is unbounded across iterations.

Per-iteration cost stays bounded (the eligible set itself is small), but:
- `presence/changed` fan-out continues to send events for agents the watcher no longer cares about
- Memory grows with `O(agents-ever-watched)` not `O(agents-currently-watched)`
- The arena would have no way to express "I no longer care about this contact's presence" without dropping the WS connection (not a reasonable v1 ergonomics)

This is a real correctness gap, not just a perf wart — the watcher's intent is "watch this set right now" and the protocol doesn't let it express that.

## Proposed fix (two options)

**Option A (preferred):** Replace-semantics on `presence/subscribe`. The connection's full subscriber set is REPLACED with `params.agentIds`. To stop watching, pass an empty array. Matches the SDK's `subscribe({filter})` pattern at `packages/client/src/runtime/subscribers.ts` — one subscription per concern, replace to update, single source of truth.

Migration: existing single-call usages (boot-time subscribe + never change) keep working because there's no second call. Existing multi-call usages (none I could find in the upstream test suite or the arena) would need to compute the union themselves before calling. That's a stricter contract but matches what every long-running client actually wants.

**Option B:** Add `presence/unsubscribe({agentIds})` RPC. Removes connId from each named agent's subscriber set. Strictly additive — no breaking change. Worse ergonomics (callers maintain their own previous-set bookkeeping to compute the diff), but safer to land.

I'd ship A — the additive semantics of the current API are the bug, not the feature.

## Downstream impact

After this lands, arena's BYOA watcher does one thing per iteration:
```typescript
yield* presence.subscribe({ agentIds: currentEligibleSet })
```

and the previous-set is implicitly cleared. ~3 LoC simpler than computing the diff manually.

## Cross-references

- #481 — `agents/list` contact-scoping (filed today; same review pass; this issue is the corollary on the presence subsystem)
- chughtapan/moltzap-arena BYOA migration plan — downstream consumer

## Source

`/plan-eng-review` of `chughtapan/moltzap-arena` BYOA migration, 2026-05-05. Outside-voice review by Codex flagged the accumulation; verified by reading `services/presence.service.ts:51-58` and `task/handlers/presence.handlers.ts:24-32`. Confirmed there is no client-side unsubscribe API in `packages/protocol/src/task/methods/presence.ts`.

chughtapan:
Closes #487. Hard blocker for arena#342 (BYOA migration epic).

`PresenceService.subscribe(connId, agentIds)` at `packages/server/src/services/presence.service.ts@69933d0` was append-only — each call merged `connId` into the union of every `agentIds` set ever passed, with no per-call un-subscription. Long-running clients that re-evaluate which agents they watch (canonical "live address book" pattern) leaked fan-out monotonically. Per Option A in #487, this PR changes the semantics to per-call replace: after the call, `connId` is in the subscriber set for EXACTLY the agents in `agentIds`, and only those. `subscribe([])` is now a functional unsubscribe-all.

## What changed

- `PresenceService` tracks a private per-connection subscription set so `subscribe()` can compute the diff without scanning every agent's subscriber set. On call: remove `connId` from each previously-watched agent absent from the new set, idempotently add to each agent in the new set, persist the new set (or delete the entry on empty).
- `removeConnection()` also clears the per-conn entry to keep the two maps in sync.
- Protocol method docstring on `PresenceSubscribe` documents replace-semantics for client authors.
- Wire RPC schema unchanged: `presence/subscribe({ agentIds })`. Handler unchanged.

## Plan anchors

| Change | Anchor |
|---|---|
| `subscribe()` becomes replace-set | #487 "Proposed fix Option A" + team-lead acceptance §1-2 |
| `connSubscriptions` private map | #487 "Implementation: replace ... in one atomic step" |
| `removeConnection()` clears per-conn entry | invariant: per-conn map mirrors `subscribers` membership |
| Protocol docstring on `PresenceSubscribe` | team-lead acceptance §6 ("document semantic shift in protocol method docstring") |
| 5 new unit tests (`presence.service.test.ts`) | team-lead acceptance §3 (the two named tests + replace-semantics edge cases) |

## Tests

Unit tests added (5 new, pinning replace-semantics):
- `subscribe replaces the prior set — connId removed from agents not in the new set` — the canonical `subscribe([A,B])` then `subscribe([B,C])` → A=∅, B=yes, C=yes test from the brief
- `subscribe replace-semantics does not disturb other connections' subscriptions` — replace on connA leaves connB's set intact
- `subscribe([]) unsubscribes the connection from all agents` — the unsubscribe-all test from the brief
- `subscribe is idempotent when the set is unchanged` — re-subscribing the same set is a no-op
- `subscribe after subscribe([]) re-establishes subscriptions cleanly` — empty doesn't poison future calls

Existing 8 `PresenceService` tests all still pass; the integration test at `26-presence-lifecycle.integration.test.ts` is wire-level and unaffected (single-call subscribe).

## Pre-PR gates

All seven green at `69933d0`:
- `pnpm typecheck` ✅
- `pnpm lint` ✅ (0 warnings, 0 errors)
- `pnpm format:check` ✅
- `bash packages/protocol/scripts/check-divergence-proofs.sh` ✅ (26 proof-required, 28 legacy exemptions)
- `pnpm test` ✅ (server: 230 unit tests pass, presence.service.test.ts: 13/13 — was 8 before)
- `pnpm test:integration` ✅ (server: 96 pass + 3 todo)
- `pnpm --filter @moltzap/client test:conformance` ✅ (10 tests pass)

## Scope

- Modules touched: `packages/server/src/services/presence.service.ts`, `packages/server/src/services/presence.service.test.ts`, `packages/protocol/src/task/methods/presence.ts`
- Tier: senior (3 files, +106 −1)
- New modules: 0
- New public signatures: 0 (signature of `subscribe()` unchanged)
- New deps: 0
- Wire schema change: 0

## Phase 12 coordination

Phase 12 (protocol layered finalization, in flight on `staff/452-phase-12-protocol-finalization`) does not touch `presence.service.ts` or the `presence/subscribe` schema per its sub-issue acceptance. If Phase 12 lands first and surfaces a conflict, will rebase.

## Confidence

HIGH — the replace-semantics is the smallest possible behavior change that satisfies the issue; new tests pin the specific scenarios from the brief; all seven pre-PR gates pass.

### DIFF
--- a/packages/protocol/src/task/methods/presence.ts
+++ b/packages/protocol/src/task/methods/presence.ts
@@ -15,6 +15,14 @@ export const PresenceUpdate = defineRpc({
   result: Type.Object({}, { additionalProperties: false }),
 });
 
+/**
+ * Subscribe to presence updates for a set of agents. Replace-semantics:
+ * the call replaces the connection's full subscriber set with
+ * `agentIds`. Agents previously subscribed to but absent from `agentIds`
+ * are silently removed; passing an empty array unsubscribes from all.
+ * Long-running clients that re-evaluate their watch set per iteration
+ * can call this idempotently without leaking fan-out.
+ */
 export const PresenceSubscribe = defineRpc({
   name: "presence/subscribe",
   params: Type.Object(
--- a/packages/server/src/services/presence.service.test.ts
+++ b/packages/server/src/services/presence.service.test.ts
@@ -136,4 +136,66 @@ describe("PresenceService", () => {
       "c-w2",
     ]);
   });
+
+  it("subscribe replaces the prior set — connId removed from agents not in the new set (#487)", () => {
+    const r = recordingSink();
+    const service = new PresenceService(r.sink);
+
+    service.subscribe("c-watcher", ["agent-a", "agent-b"]);
+    service.subscribe("c-watcher", ["agent-b", "agent-c"]);
+
+    expect([...service.getSubscribers("agent-a")]).toEqual([]);
+    expect([...service.getSubscribers("agent-b")]).toEqual(["c-watcher"]);
+    expect([...service.getSubscribers("agent-c")]).toEqual(["c-watcher"]);
+  });
+
+  it("subscribe replace-semantics does not disturb other connections' subscriptions (#487)", () => {
+    const r = recordingSink();
+    const service = new PresenceService(r.sink);
+
+    service.subscribe("c-w1", ["agent-a", "agent-b"]);
+    service.subscribe("c-w2", ["agent-a", "agent-b"]);
+    service.subscribe("c-w1", ["agent-b"]);
+
+    expect([...service.getSubscribers("agent-a")]).toEqual(["c-w2"]);
+    expect([...service.getSubscribers("agent-b")].sort()).toEqual([
+      "c-w1",
+      "c-w2",
+    ]);
+  });
+
+  it("subscribe([]) unsubscribes the connection from all agents (#487)", () => {
+    const r = recordingSink();
+    const service = new PresenceService(r.sink);
+
+    service.subscribe("c-watcher", ["agent-a", "agent-b", "agent-c"]);
+    service.subscribe("c-watcher", []);
+
+    expect([...service.getSubscribers("agent-a")]).toEqual([]);
+    expect([...service.getSubscribers("agent-b")]).toEqual([]);
+    expect([...service.getSubscribers("agent-c")]).toEqual([]);
+  });
+
+  it("subscribe is idempotent when the set is unchanged (#487)", () => {
+    const r = recordingSink();
+    const service = new PresenceService(r.sink);
+
+    service.subscribe("c-watcher", ["agent-a", "agent-b"]);
+    service.subscribe("c-watcher", ["agent-b", "agent-a"]);
+
+    expect([...service.getSubscribers("agent-a")]).toEqual(["c-watcher"]);
+    expect([...service.getSubscribers("agent-b")]).toEqual(["c-watcher"]);
+  });
+
+  it("subscribe after subscribe([]) re-establishes subscriptions cleanly (#487)", () => {
+    const r = recordingSink();
+    const service = new PresenceService(r.sink);
+
+    service.subscribe("c-watcher", ["agent-a"]);
+    service.subscribe("c-watcher", []);
+    service.subscribe("c-watcher", ["agent-b"]);
+
+    expect([...service.getSubscribers("agent-a")]).toEqual([]);
+    expect([...service.getSubscribers("agent-b")]).toEqual(["c-watcher"]);
+  });
 });
--- a/packages/server/src/services/presence.service.ts
+++ b/packages/server/src/services/presence.service.ts
@@ -17,6 +17,10 @@ const EMPTY_SUBSCRIBERS: ReadonlySet<string> = new Set();
 export class PresenceService {
   private statuses = new Map<string, PresenceStatus>();
   private subscribers = new Map<string, Set<string>>();
+  // Per-connection record of which agentIds the connection is subscribed
+  // to. Tracked so `subscribe()` can replace the prior subscription set
+  // atomically without scanning every agent's subscriber set.
+  private connSubscriptions = new Map<string, Set<string>>();
 
   constructor(private readonly eventSink: PresenceEventSink) {}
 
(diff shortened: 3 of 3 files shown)


chughtapan (commented):
**REVIEW SENIOR — VERDICT: APPROVE_WITH_CONCERNS**

Single-pass review of #490 against #487 acceptance + team-lead brief.

## Scope

- Claimed modality: `implement-senior` (replace-semantics on `PresenceService.subscribe`).
- Observed shape: 3 files, +106/−1, 1 service module + 1 protocol docstring + 1 test file. No new public surface, no new deps, wire schema unchanged.
- Match: yes. The diff stays in lane.

## Acceptance walk

| AC | Status | Evidence |
|---|---|---|
| `subscribe(connId, agentIds)` is replace-set, not append-set | ✅ | `presence.service.ts:67-95` — diffs prev vs next set, removes from agents not in next |
| After call: `connId` in subscriber set for EXACTLY agents in `agentIds` | ✅ | `presence.service.ts:71-88` removal-then-addition; pinned by test at L140-150 |
| Removed from agents NOT in `agentIds` | ✅ | `presence.service.ts:72-78` |
| Atomic single-step (no observable mid-state) | ✅ | function body is fully synchronous; no `await` / no fiber yield between the remove-loop and the add-loop |
| Unit test: subscribe([A,B]) then [B,C] → only B,C | ✅ | `presence.service.test.ts:140-150` |
| Unit test: subscribe([]) → all removed | ✅ | `presence.service.test.ts:167-177` |
| Integration test exercises new behavior end-to-end | ⚠ deferred | PR body acknowledges existing `26-presence-lifecycle` is single-call; no multi-call wire test added. See concern below — filed #491. |
| Wire schema `presence/subscribe({ agentIds })` unchanged | ✅ | `protocol/src/task/methods/presence.ts:26-36` — only docstring added |
| Protocol docstring documents semantic shift | ✅ | `protocol/src/task/methods/presence.ts:18-25` |
| No new public surface | ✅ | only private `connSubscriptions` map added; no exports changed |
| No Phase 12/13/14 absorption | ✅ | diff confined to presence subsystem |

## Craft principles (P1–P4)

- **P1 (types beat tests):** N/A — no new constraints to encode. Existing `string` parameters on `subscribe` carry over; branding `connId`/`agentId` is out of scope for this PR.
- **P2 (boundaries):** unchanged — protocol handler decodes via existing TypeBox `agentIds: Type.Array(AgentId)` schema; no new ingress.
- **P3 (typed errors):** clean — no throws, no catches, function returns `void` and has no failure modes (in-memory map ops).
- **P4 (exhaustiveness):** N/A — no new switches, no new union pattern matching. The `if (next.size === 0)` branch terminates explicitly with else.

No craft findings.

## Tests

5 new unit tests (`presence.service.test.ts:140-200`) pin the replace-semantics:
- replace removes prior agents not in new set
- replace doesn't disturb other connections (cross-conn isolation)
- `subscribe([])` clears all
- idempotent when set unchanged
- `subscribe([])` doesn't poison subsequent calls

Existing 8 `PresenceService` tests still pass per PR body. CI green: ci ✅, conformance ✅, test ✅.

The unit-test set is exactly what the brief named, plus two replace-edge cases (idempotence + post-empty re-subscribe). Coverage at the service boundary is comprehensive.

## Concerns (deferrable to follow-up)

**1. End-to-end integration test for replace-semantics — filed as #491.**

PR body states: "the integration test at `26-presence-lifecycle.integration.test.ts` is wire-level and unaffected (single-call subscribe)." That's accurate — no integration test exercises the multi-call replace path. End-to-end coverage today is by composition: wire→handler routing (existing integration tests), handler→service.subscribe is a one-line pass-through at `task/handlers/presence.handlers.ts:28`, service.subscribe replace-semantics (5 new unit tests). The composition is sound, so this is a regression-suite gap, not a correctness gap.

Filed #491 for a multi-call wire-level test (`subscribe([A,B])` → `subscribe([B,C])` → assert no `presence/changed` for A reaches the watcher). **Deferrable to follow-up — does not block merge.** A future refactor that breaks the handler pass-through would not be caught by today's wire suite, but the present diff doesn't introduce that risk.

## Notes (non-blocking, not flagged as concerns)

- Empty `Set` entries linger in `this.subscribers` after a replace removes the last subscriber for an agent (the map grows with `O(agents-ever-subscribed-to)`). This is the same shape as the pre-existing `removeConnection` path — not regressed by this PR. Out of scope.
- `connSubscriptions.set(connId, next)` stores the local `Set` directly; readers of the map don't mutate it, so no aliasing leak. Fine.

## Verdict reasoning

The implementation is the smallest possible behavior change that satisfies #487. Replace-semantics is correct; the diff between prev and next is computed before mutation, removal happens before addition, and the per-conn map is kept in sync with `subscribers` membership on `removeConnection`. Unit tests pin every named scenario from the brief plus two edge cases. Wire schema and handler are unchanged.

Verdict is `APPROVE_WITH_CONCERNS` — concern is one deferrable follow-up (#491) for end-to-end regression-suite coverage. Per the contract amendment, the concern is filed as a GH issue, not just listed here. Per `feedback_read_concerns_before_merging.md`: this concern is **deferrable to follow-up issue**, not resolve-before-merge. The orchestrator may proceed to land #490.

Routing: dispatch `/safer:verify` next on the closed criteria; #491 is its own implement-junior dispatch when prioritized.

PR #493, opened 2026-05-06
[impl-staff] Phase 12 expanded: JSON-RPC layered finalization (#452)

chughtapan:
Closes #452. Plan: [docs/plans/phase-12-jsonrpc-finalization-2026-05.md](docs/plans/phase-12-jsonrpc-finalization-2026-05.md).

## Status

Squash `37e74f3` lands S1–S12.1; CI 3/3 green (ci, test, conformance); ready for senior review.

## Reviewer's next step

Start at [`packages/protocol/src/index.ts`](https://github.com/chughtapan/moltzap/blob/37e74f3/packages/protocol/src/index.ts) (54% smaller surface, just `export *` re-exports from per-layer barrels). Then [`packages/protocol/src/transport/json-rpc-client.ts`](https://github.com/chughtapan/moltzap/blob/37e74f3/packages/protocol/src/transport/json-rpc-client.ts) and [`packages/protocol/src/transport/json-rpc-server.ts`](https://github.com/chughtapan/moltzap/blob/37e74f3/packages/protocol/src/transport/json-rpc-server.ts) (new originator/responder primitives). Skim per-scope files in `identity/` (`agents.ts`, `contacts.ts`, `invites.ts`) and `task/` (`conversations.ts`, `messages.ts`, `tasks.ts`) — each owns its brands + errors + schemas + RPCs + notifications.

## Wire-method renames

Phase 12 reshuffles namespaces; consumers must update callsites:

| Before | After | Notes |
|---|---|---|
| `auth/connect` | `network/connect` | Connection-handshake; lives in network layer |
| `auth/register` | `agents/register` | Registration is an agent operation |
| `auth/invite-agent` | `agents/invite` | |
| `auth/claim` | `agents/claim` | New in #486; renamed for namespace consistency |
| `system/ping` | `network/ping` | Health check |
| `auth/select-agent` | _DELETED_ | Dead surface (no consumer post-Phase 8) |

## Layer ownership

| Layer | Owns |
|---|---|
| `transport/` | Wire frames (request/response/notification), `JsonRpcClient` + `JsonRpcServer` runtime primitives, `handler<D, Ctx>(def, fn)` factory, JSON-RPC reserved-code tagged errors (`Unauthorized`, `Forbidden`, `NotFound`, `Conflict`, `InvalidParams`), `MalformedFrameError`, `registerErrorClass` (internal), error-class registry |
[...]

PR #505, opened 2026-05-07
fix: contact-scope agents/list + presence/subscribe (#481)

### LINKED ISSUE #481: agents/list returns all server agents — should be contact-scoped
Surfaced from a downstream consumer (`moltzap-arena` BYOA migration) during `/plan-eng-review` on 2026-05-05.

## What

`agents/list({})` (defined at `packages/protocol/src/network/methods/auth.ts:155-160`, returns `Record<AgentId, AgentCard>`) currently returns **every registered agent on the server**, irrespective of whether the caller's owner has any relationship to the listed agents' owners. The new docs confirm the intent: "List all registered agents on the server" (`docs/protocol/methods/agents-list.mdx`).

Same concern applies to `presence/subscribe({agentIds})` — it accepts arbitrary agent IDs and returns presence for them, with no check that the caller's owner is allowed to see those agents at all.

## Why this is a bug, not a feature

**1. Policy-incoherent with the contact gate.** Commit `ff93bbe` (closes #361) made direct-conversation creation gated on `ContactService.areInContact(initiatorOwner, targetOwner)`. So the protocol's stance is "you can't *message* an agent whose owner you're not in contact with" — but the same caller can freely *enumerate* every agent on the server. The directory leaks exactly what the messaging gate is trying to protect. If a caller can `agents/list` to see every agent, the contact gate is just a delay, not a policy.

**2. Privacy leak.** Any registered agent — and registration is intentionally easy (auth/register with optional invite code) — can enumerate the entire server's agent registry: agent IDs, names, `ownerUserId`, descriptions, tags. There's no opt-in.

**3. Scales unboundedly.** Response size is O(all-agents-ever-registered). Fine at single-tenant v1; a wall at v2.

**4. The previous scoping (pre-Phase-7) was conversation-scoped, which had its own bug.** Downstream arena's existing watcher carries this comment as evidence:

> Why `agents/lookupByName` and not `agents/list`: the network's `agents/list` is conversation-scoped (only returns agents the caller already shares a conversation with). At boot, arena-gm has no sessions, so `agents/list` would return `{}` forever — chicken and egg. (chughtapan/moltzap-arena: `packages/server/src/werewolf-app/agent-presence-watcher.ts:13-19`)

So agents/list was scoped, just by the wrong primitive (conversations, which are downstream of the right primitive). Phase 7 removed conversation-scoping; instead of substituting the right scope, upstream went fully-public. That's an over-correction.

## Proposed semantic

`agents/list({includeSelf?: boolean})`:

- **Default** (no params, or `includeSelf: false`): returns AgentCards for agents whose `ownerUserId` is an **accepted-status contact** of the caller's `ownerUserId`. **Does not include the caller's own agents.** This is the "agents I can play with / message that aren't me" set — the common case for BYOA-style discovery (arena Werewolf, escape-room, roast battle, anything with a "bring your agent" loop).

- **Opt-in** (`includeSelf: true`): also includes agents whose `ownerUserId` equals the caller's own `ownerUserId`. For sibling-discovery use cases where a user runs several agents that need to find each other.

The default-exclude-self decision matters: discovery callers don't typically need to discover themselves (they already have references), and including self forces every downstream consumer to write a `dropSelfAgent` belt-and-suspenders filter that the upstream policy should own. Concrete reference: arena's existing `dropGmAgent` filter at `chughtapan/moltzap-arena packages/server/src/werewolf-app/agent-presence-watcher.ts:184-194` exists exactly because of this gap.

`presence/subscribe({agentIds, includeSelf?: boolean})` applies the same rule: rejects (or silently drops from the subscription) agent IDs not in the caller's visible set, computed the same way. Same default-exclude-self.

`agents/lookup(agentIds)` and `agents/lookupByName(names)` — those are dereference-by-known-key, so the privacy concern is at the enumeration verb, not the lookup verb. They could keep current semantics, OR also become contact-scoped for symmetry. Lower-priority sub-decision.

Caller calling agents/list before any contacts are accepted gets `{}` (with default exclude-self), or just their own agents (with `includeSelf: true`). That's correct — "no one in the address book yet" should return no one in the directory.

## Why this matters for downstream apps

With agents/list properly contact-scoped and self-excluded by default, BYOA-style apps collapse their discovery flow to:

```
agents/list({})            // contact-scoped, self-excluded — exactly the playable population
presence/subscribe(IDs)    // gated to the same set, replace-semantics per #487
LRP-pick K when |online| >= K
tasks/create({invitedAgentIds: roster})
endpoints/registerTaskManager(task.id)
```

No client-side join of contacts/list + agents/list. No reimplementing the contact policy in app code. No `dropSelfAgent` workaround. No need for the original-draft RPCs (`contacts/listAgents`, `presence/subscribeContacts`) the downstream considered before this insight.

## Backwards compatibility

This is a behavior tightening — current callers who get back "all server agents" will get a smaller set after the fix. Most legitimate consumers should be fine since they were probably filtering client-side anyway. Worth flagging in CHANGELOG as a breaking-but-correct change. Conformance test 29-agents-list (`packages/server/src/__tests__/integration/29-agents-list.integration.test.ts`) needs updating to seed contacts before assertions.

The opt-in `includeSelf` flag preserves the sibling-discovery use case for callers that need it without making it the default footgun.

## Cross-references

- #361 — contact-policy gate on conversations (the existing primitive this issue extends to discovery)
- #474 — contact/expired notification for orphaned pending (filed earlier today; orthogonal but pairs naturally for the BYOA story)
- #486 — programmatic claim flow (filed today; pairs with this issue — once agents have proper claimed owners, contact-scoping has the data it needs)
- #487 — presence/subscribe replace-semantics (filed today; corollary on the presence subsystem — pairs with this issue's `presence/subscribe` scoping)

## Source

- /plan-eng-review of `chughtapan/moltzap-arena` BYOA migration, 2026-05-05
- Outside-voice (Codex) pass during the review surfaced the self-exclude amendment after the initial filing
- Cross-referenced from arena issue chughtapan/moltzap-arena#336 (contact-storm rate limiting) for the broader BYOA discovery design conversation

chughtapan:
Closes #481.

## Summary

`agents/list({})` and `presence/subscribe({agentIds})` are now contact-scoped against the caller's `ownerUserId`. The visible set is:

- the caller's own `agentId` (always — unclaimed callers at least see themselves);
- agents owned by the caller's `ownerUserId` (sibling agents);
- agents owned by an `accepted`-status contact of the caller's `ownerUserId`.

`agents/list` previously SQL-joined on `conversation_participants`, leaking the entire registry once any conversation existed (and pre-Phase-7 left fresh callers with `{}` until they shared a conversation — chicken-and-egg for BYOA). `presence/subscribe` previously accepted arbitrary IDs without any permission check; now it silently drops agentIds outside the visible set. **Wire schemas unchanged** — filter logic moved into the handler.

Unblocks **arena#342** BYOA discovery: clients can now call `agents/list({})` + `presence/subscribe(IDs)` and get exactly the playable population without re-implementing the contact policy in app code.

## Implementation

A single shared helper at `packages/server/src/services/agent-visibility.ts` (`visibleAgentIds(...)`) is the source of truth for both handlers. The Kysely contact-graph join + `restrictTo` intersection live in one place; both `agents/list` (no `restrictTo`) and `presence/subscribe` (intersect with `params.agentIds`) route through it.

## Tests

- **Unit** — `services/agent-visibility.test.ts` (PGlite, 7 cases): own + sibling include, accepted contact include, pending-status excluded, `restrictTo` intersection, empty `restrictTo` short-circuit, unclaimed self-only with and without `restrictTo`.
- **Integration (rewritten)** — `__tests__/integration/29-agents-list.integration.test.ts`: cross-owner via admin-register + `contacts/add` + `contacts/accept`. Covers AC bullet #3: A in-contact-with B → A's `agents/list` returns A's + B's, never C's; pending-only relationship does NOT unlock visibility; `AgentCard` fields propagate correctly through the contact-visible path.
- **Integration (preserved)** — `__tests__/integration/26-presence-lifecycle.integration.test.ts`: now seeded with one shared `devModeUserId` so the broadcast-lifecycle assertions stay in scope without needing per-test contact setup.
- **Conformance** — server-side conformance ($SKIP_TOXIPROXY=1$, 24 passed / 5 deferred / 6 unavailable) and client-side conformance (10 passed) both green; the conformance suite already runs with `devModeUserId` so its presence properties remain valid under the new scope.

## Pre-PR gates

| Gate | Result |
| --- | --- |
| `pnpm typecheck` | ✅ |
| `pnpm lint` | ✅ (2 pre-existing warnings, not from this diff) |
| `pnpm format:check` | ✅ |
| `bash packages/protocol/scripts/check-divergence-proofs.sh` | ✅ |
| `pnpm test` | ✅ 28 files / 227 tests |
| `pnpm test:integration` | ✅ 31 files / 105 tests |
| `pnpm --filter @moltzap/client test:conformance` | ✅ 10 properties pass |
| `pnpm --filter @moltzap/server-core test:conformance` (no-toxic) | ✅ 24 pass / 5 deferred |

## Branch base

Branched off PR #493 (`staff/452-phase-12-protocol-finalization`). GitHub will auto-retarget to `main` on Phase-12 merge.

### DIFF
--- a/packages/server/src/__tests__/integration/26-presence-lifecycle.integration.test.ts
+++ b/packages/server/src/__tests__/integration/26-presence-lifecycle.integration.test.ts
@@ -16,8 +16,12 @@ import {
   PresenceChangedNotificationDefinition,
 } from "@moltzap/protocol";
 
+// Sibling-owner cohort so #481 contact-scoping doesn't filter agents
+// out — this file exercises broadcast lifecycle, not the visibility gate.
+const PRESENCE_DEV_OWNER = "00000000-0000-4000-8000-000000000470";
+
 beforeAll(async () => {
-  await startTestServer();
+  await startTestServer({ devModeUserId: PRESENCE_DEV_OWNER });
 }, 60_000);
 
 afterAll(async () => {
--- a/packages/server/src/__tests__/integration/29-agents-list.integration.test.ts
+++ b/packages/server/src/__tests__/integration/29-agents-list.integration.test.ts
@@ -5,29 +5,40 @@ import {
   startTestServer,
   stopTestServer,
   resetTestDb,
-  registerAndConnect,
   getKyselyDb,
   trackClient,
-  registerAgent,
   connectTestClient,
 } from "./helpers.js";
 import type { AgentCard } from "@moltzap/protocol";
+import { userId } from "@moltzap/protocol/testing";
+import type { UserId } from "@moltzap/protocol/identity";
 
 import {
   AgentsList,
   AgentsLookup,
   AgentsLookupByName,
-  ConversationsCreate,
+  ContactsAdd,
+  ContactsAccept,
 } from "@moltzap/protocol";
 
 type AgentsListResult = { agents: Record<string, AgentCard> };
 type AgentsArrayResult = { agents: AgentCard[] };
 
+// agents/list is contact-scoped per #481; admin-register is used to bind
+// explicit owners so the cross-owner visibility cases can be exercised.
+const REGISTRATION_SECRET = "agents-list-test-secret-zxcv";
+const ALICE_USER_ID = userId("00000000-0000-4000-8000-00000000a11c");
+const BOB_USER_ID = userId("00000000-0000-4000-8000-00000000b0b0");
+const CAROL_USER_ID = userId("00000000-0000-4000-8000-00000000ca60");
+
 let baseUrl: string;
 let wsUrl: string;
+let pairCounter = 0;
 
 beforeAll(async () => {
-  const server = await startTestServer();
+  const server = await startTestServer({
+    registrationSecret: REGISTRATION_SECRET,
+  });
   baseUrl = server.baseUrl;
   wsUrl = server.wsUrl;
 }, 60_000);
@@ -38,12 +49,61 @@ afterAll(async () => {
 
 beforeEach(async () => {
   await resetTestDb();
+  pairCounter = 0;
 });
 
-/** Register an agent with custom options (e.g. description), tracked for cleanup. */
-function registerWithOpts(name: string, opts: { description?: string }) {
+interface AdminRegisterResponse {
+  agentId: string;
+  apiKey: string;
+}
+
+async function adminRegister(
+  name: string,
+  ownerUserId: UserId,
+  description?: string,
+): Promise<AdminRegisterResponse> {
+  const res = await fetch(`${baseUrl}/api/v1/admin/register-agent`, {
+    method: "POST",
+    headers: { "Content-Type": "application/json" },
+    body: JSON.stringify({
+      name,
+      inviteCode: REGISTRATION_SECRET,
+      ownerUserId,
+      ...(description !== undefined ? { description } : {}),
+    }),
+  });
+  const json = (await res.json()) as AdminRegisterResponse;
+  if (res.status !== 201 && res.status !== 200) {
+    throw new Error(
+      `admin register failed: ${res.status} ${JSON.stringify(json)}`,
+    );
+  }
+  return json;
+}
+
+interface OwnedConnectedAgent {
+  agentId: string;
+  ownerUserId: UserId;
+  client: Awaited<
+    ReturnType<typeof connectTestClient> extends Effect.Effect<
+      infer A,
+      infer _E
+    >
+      ? A
+      : never
+  >;
+}
+
+function registerAndConnectOwned(opts: {
+  name: string;
+  ownerUserId: UserId;
+  description?: string;
+}): Effect.Effect<OwnedConnectedAgent, Error> {
   return Effect.gen(function* () {
-    const reg = yield* registerAgent(baseUrl, name, opts);
+    const idx = ++pairCounter;
+    const reg = yield* Effect.tryPromise(() =>
+      adminRegister(`${opts.name}-${idx}`, opts.ownerUserId, opts.description),
+    );
     const client = yield* connectTestClient({
       wsUrl,
       agentId: reg.agentId,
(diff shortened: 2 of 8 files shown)


chughtapan (commented):
**REVIEW SENIOR — VERDICT: APPROVE_WITH_CONCERNS**

Single-pass review, ~35 min, 8-file diff (+597/-166). Security-adjacent fix; the core change closes a real privacy leak and the implementation is clean.

## Acceptance walk (#481)

| AC | Status | Evidence |
|---|---|---|
| `agents/list({})` filters by caller contact graph + caller's own agents (self-include) | ✅ | `auth.handlers.ts:215-244` routes through `visibleAgentIds`; `agent-visibility.ts:48-62` SQL: own + sibling + accepted-contact-owned |
| `presence/subscribe({agentIds})` filters out-of-contact agentIds | ✅ | `presence.handlers.ts:27-43` intersects `params.agentIds` with `visibleAgentIds(restrictTo)`; subscribe + getMany both use filtered set |
| Unit: A in-contact-with B → A's `agents/list` returns A's + B's agents (not C's) | ✅ | `agent-visibility.test.ts:107-124` |
| Unit: A not-in-contact-with C → A's `presence/subscribe([C])` errors or filters | ✅ | `agent-visibility.test.ts:150-168` (helper-level intersection test) |
| Integration: register → contacts/add → contacts/accept → A's agents/list shows B's agents | ✅ | `29-agents-list.integration.test.ts:270-313` |
| Wire schemas unchanged | ✅ | `protocol/identity/methods.ts:193-200` AgentsList unchanged; `presence/subscribe` schema unchanged |
| Pending-status contacts excluded | ✅ | `agent-visibility.test.ts:126-148` + `29-agents-list.integration.test.ts:332-361` |
| Unclaimed callers see only themselves | ✅ | `agent-visibility.ts:32-39` short-circuit + 2 unit tests |

## arena#344 BYOA unblock — verified

A fresh GM with one accepted contact can discover that contact's agents: confirmed by the integration test "returns agents whose owner is an accepted contact" (Alice + accepted-Bob → Alice's `agents/list` includes Bob's agents; Carol who is no contact is not visible). The `dropGmAgent` filter in arena's watcher will need to stay (because the PR includes-self always), which is a deviation from #481's "default-exclude-self" — but matches team-lead's assignment that says "self-include." Documented divergence, not a defect.

## Local gates re-run

- `pnpm typecheck` ✅ (no errors)
- `pnpm --filter @moltzap/server-core test` ✅ (28 files / 227 tests, including new `agent-visibility.test.ts` 7 cases)

## Concerns (non-blocking — filed as follow-up issues)

1. **#506** — `agents/lookup` + `agents/lookupByName` remain unscoped. #481 named this a "lower-priority sub-decision"; `lookupByName` is the bigger leak (32-char dictionary attacks tractable). The new `visibleAgentIds(restrictTo)` helper is the obvious reuse point.
2. **#507** — `scripts/generate-protocol-docs.ts:79-81` (`agents/list`) and `:203-205` (`presence/subscribe`) still describe pre-PR behavior. Stale text propagates to `docs/protocol/methods/{agents-list,presence-subscribe}.mdx`. Should be a one-shot update + regenerate.
3. **#508** — `presence/subscribe` silent-drop is allowed by #481 but creates a discoverability gap (caller can't distinguish "offline" from "policy-filtered"). Tagged-error or `droppedAgentIds` in result; review-senior craft note prefers the former.

## Craft-quality notes

- Helper has clean separation (single source of truth, two callers, tested behavior at the helper level).
- `restrictTo === undefined ? base : base.where("id", "in", restrictTo as AgentId[])` is the right shape for a reusable visibility predicate.
- Unclaimed (`callerOwnerUserId === null`) early-returns are exhaustive over the only two states; no DB hit on the unclaimed path.
- `restrictTo as AgentId[]` and `ids as ServerAgentId[]` casts are noise (same brand under the hood) but consistent with existing patterns in this file — not worth fixing in this PR.
- `26-presence-lifecycle.integration.test.ts` `devModeUserId` workaround is the right minimal change to keep broadcast-lifecycle assertions in scope without per-test contact dance; well-commented.
- `requiresActive: true` on `agents/list` is preserved; unclaimed-but-active agents (rare but possible) hit the `null`-owner branch and see only themselves. Correct.

## Verdict

**APPROVE_WITH_CONCERNS** — security-positive, well-tested, helper is the right shape. Three follow-up issues filed (#506, #507, #508) for items I'd otherwise downgrade verdict for. Safe to merge after Phase-12 (#493) lands and base auto-retargets.

— review-senior, 2026-05-07

PR #517, opened 2026-05-08
[impl-junior] fix(protocol): json-rpc-client preserves data field on registered error decode (#511)

### LINKED ISSUE #511: fix(protocol): json-rpc-client drops data field when deserializing registered error classes
## Problem

`packages/protocol/src/transport/json-rpc-client.ts` constructs registered error class instances with `new cls()` (no arguments) when decoding an inbound error frame whose code matches the wire-error registry:

```typescript
const failureValue: RpcCallError =
  cls === undefined
    ? new RpcServerError({ code: frame.error.code, message: frame.error.message, data: frame.error.data })
    : (new cls() as RegisteredTaggedError);  // data from frame.error.data is dropped
```

Any registered error that carries a `data` payload (e.g., `NotInContactsError.data.agentIds` added in #510) arrives with `data: undefined` on the client side for consumers using `makeJsonRpcClient` (the server→client app-callback channel).

## Impact

Currently limited to the server→client app-callback channel (`packages/server/src/ws/connection.ts`). The `presence/subscribe` path (client→server) uses `TestClient.sendRpc` which constructs `RpcResponseError` directly from `frame.error.data`, so #508/#510 is not affected.

If future server→client RPCs fail with a registered error that carries a `data` payload, the payload will be silently lost.

## Fix

Pass the wire frame's `data` to the constructor:

```typescript
: (new cls({ data: frame.error.data } as never) as RegisteredTaggedError);
```

Or extend `RpcErrorClass` to accept a factory method that takes `{ message?, data? }` so the typed constructor signature is honoured rather than cast.

## Source

Surfaced during review of PR #510 (`packages/protocol/src/transport/json-rpc-client.ts`, `resolve` function).

chughtapan:
Closes #511. Refs #512.

## What changed

`packages/protocol/src/transport/json-rpc-client.ts` (`resolve` function): when an inbound error frame's code matched the wire-error registry, the registered tagged-error class was constructed with `new cls()` and the wire `data` payload was silently dropped. The fix forwards `frame.error.data` to the constructor so any payload set by the server reaches the typed instance rather than arriving as `undefined`.

Adds `packages/protocol/src/transport/json-rpc-client.test.ts` — first direct unit test for `makeJsonRpcClient`. Covers the registered-class decode branch via `ForbiddenError` with a `{ agentIds: [...] }` data payload, asserting the failure reaches the call site as a typed `ForbiddenError` instance with `data` preserved.

## Scope

- Module: `packages/protocol/src/transport/json-rpc-client.{ts,test.ts}`
- Tier: junior (one-module fix + regression test)
- New exported signatures: none
- New deps: none
- The `as never` cast retained — forced by `RpcErrorClass`'s open `...args: never[]` constructor signature in `wire-errors.ts:19-21`. A typed factory hook on `RpcErrorClass` would localize the unsafety; filing as a follow-up issue rather than expanding scope here.

## Tests

- `preserves the wire frame's data payload on a registered error code` — feeds a synthetic error frame via `client.resolve`, awaits the call's failure, asserts the failure value is a `ForbiddenError` instance with the original `data` payload intact (regression guard for #511).
- Full protocol package suite: 114 passed | 1 todo (115) on `pnpm --filter @moltzap/protocol test`.

## Review skips

- `/review` (Step 4 critical pass categories): not applicable to a 5-line transport-layer object-construction fix — no SQL, LLM, shell, enum changes; no async/sync mixing (Effect-native throughout); no trust-boundary surface change (data is already on the wire, the fix forwards it).
- Three independent review passes (reuse / quality / efficiency) ran via `/simplify` against the full diff. One actionable finding applied: comment tightened to drop a misleading typed-payload example (data is `unknown` on `RpcErrorPayload`, not a structurally-typed field).
- Two follow-ups deferred to GitHub issues (filed post-merge): (a) typed factory hook on `RpcErrorClass` to remove the `as never` cast; (b) symmetric unit test for the unregistered-code branch (`RpcServerError`).

## Confidence

HIGH — fix recipe matches issue body verbatim; new test fails on prior code (verified by inspection: `new cls()` constructs with no `data` arg) and passes on new code.

## Process issues

- Recovered from a prior teammate's tmux restart; existing worktree had the diff already staged. Resumed from worktree per Amendment 2.

### DIFF
--- a/packages/protocol/src/transport/json-rpc-client.test.ts
+++ b/packages/protocol/src/transport/json-rpc-client.test.ts
@@ -1,0 +1,81 @@
+/**
+ * Unit tests for `JsonRpcClient.resolve` — focused on inbound error-frame
+ * decode for codes that match the registered tagged-error registry.
+ *
+ * The resolver path that constructs the failure value lives in
+ * `json-rpc-client.ts` and was the site of issue #511 (the wire frame's
+ * `data` payload was dropped because the registered class was constructed
+ * with no arguments).
+ */
+import { describe, expect, it } from "vitest";
+import { Cause, Deferred, Effect, Exit, Fiber } from "effect";
+import { Type } from "@sinclair/typebox";
+
+import { defineRpc } from "./method.js";
+import { makeJsonRpcClient } from "./json-rpc-client.js";
+import { responseFrame, type ResponseFrame } from "./wire.js";
+import { ForbiddenError } from "./wire-errors.js";
+
+const TestEcho = defineRpc({
+  name: "test/echo",
+  params: Type.Object({}, { additionalProperties: false }),
+  result: Type.Object({}, { additionalProperties: false }),
+});
+
+function parseRequestId(raw: string): string {
+  const parsed = JSON.parse(raw) as { id?: unknown };
+  if (typeof parsed.id !== "string") {
+    throw new Error(`expected string id in outgoing frame, got ${raw}`);
+  }
+  return parsed.id;
+}
+
+describe("JsonRpcClient.resolve — registered tagged-error decode", () => {
+  it("preserves the wire frame's `data` payload on a registered error code", async () => {
+    const wireData = { agentIds: ["agent-a", "agent-b"] } as const;
+
+    const exit = await Effect.runPromise(
+      Effect.scoped(
+        Effect.gen(function* () {
+          // Capture the request id the client mints on the first write.
+          // Effect.ignore makes a second write (none expected) a no-op.
+          const idDef = yield* Deferred.make<string>();
+          const client = yield* makeJsonRpcClient({
+            write: (raw) =>
+              Deferred.succeed(idDef, parseRequestId(raw)).pipe(Effect.ignore),
+            idPrefix: "test",
+          });
+
+          // Issue the call on a fiber so it parks on the deferred while
+          // the test feeds an inbound error frame via `resolve`.
+          const callFiber = yield* Effect.fork(client.call(TestEcho, {}));
+          const outgoingId = yield* Deferred.await(idDef);
+
+          const inbound: ResponseFrame = responseFrame(outgoingId, {
+            error: {
+              code: ForbiddenError.code,
+              message: "denied",
+              data: wireData,
+            },
+          });
+
+          const handled = yield* client.resolve(inbound);
+          expect(handled).toBe(true);
+
+          return yield* Effect.exit(Fiber.join(callFiber));
+        }),
+      ),
+    );
+
+    expect(Exit.isFailure(exit)).toBe(true);
+    if (!Exit.isFailure(exit)) return;
+    const failure = Cause.failureOption(exit.cause);
+    expect(failure._tag).toBe("Some");
+    if (failure._tag !== "Some") return;
+    const value = failure.value;
+    expect(value).toBeInstanceOf(ForbiddenError);
+    // Regression guard for #511: the wire `data` payload reaches the
+    // typed instance instead of arriving as `undefined`.
+    expect((value as ForbiddenError).data).toEqual(wireData);
+  });
+});
--- a/packages/protocol/src/transport/json-rpc-client.ts
+++ b/packages/protocol/src/transport/json-rpc-client.ts
@@ -90,8 +90,12 @@ export const makeJsonRpcClient = (config: {
                 // in `RegisteredTaggedError`. The type system can't see
                 // through the open-ended `new (...) => { _tag: string }`
                 // factory shape, so the cast bridges the static factory
-                // to the closed runtime union.
-                (new cls() as RegisteredTaggedError);
+                // to the closed runtime union. The `data` argument is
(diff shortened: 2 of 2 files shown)


chughtapan (commented):
## Verdict: APPROVE

**Aggregate verdict (single-reviewer / Amendment 4 implement-junior tier).** CI status: green (ci, conformance, test = SUCCESS; Mintlify Deployment skipped per docs path). Diff is 87/-2 across two files in `packages/protocol/src/transport/`, exactly the surface named in the issue.

### Diff matches issue #511 fix recipe

`packages/protocol/src/transport/json-rpc-client.ts:96-98@b28b4e4` replaces `new cls()` with `new cls({ data: frame.error.data } as never)` as the issue body prescribes. The accompanying inline comment is updated to document the `#511` regression guard. The `as never` cast is the recipe's named option (the alternative factory-method extension was explicitly an "or"), so this is the lighter of the two acceptable shapes.

### Test coverage

`packages/protocol/src/transport/json-rpc-client.test.ts:1-81@b28b4e4` covers the positive branch: a registered code (`ForbiddenError.code = -32001`) with a non-empty `data` payload (`{ agentIds: ["agent-a", "agent-b"] }`) round-trips through `resolve` and surfaces on the `ForbiddenError` instance. Test shape:
- Forks the call fiber so it parks on the deferred, then drives inbound through `client.resolve(...)`.
- Captures the outgoing request id via a `Deferred<string>` rather than guessing the prefix-counter format (correct insulation from the id-minting contract).
- Uses the typed `responseFrame(...)` factory rather than hand-rolled JSON, so the inbound frame is `ResponseFrame`-shaped at the type level.
- Asserts via `Cause.failureOption` after `Exit.isFailure` — instanceof + `expect(value.data).toEqual(wireData)` is the regression guard.

### Craft floor (P1-P4)

- **P1 (Types beat tests).** The `as never` cast is a known boundary between the open-ended `RpcErrorClass` factory shape and the typed payload arms. Issue body authorizes it explicitly. Cleaner P1 alternative (factory-method on `RpcErrorClass`) is deferrable; not a blocker.
- **P2 (Validate at boundaries).** Inbound test frame uses the typed `responseFrame` factory. Wire-side `data: unknown` flows through unchanged — structural to the wire-error contract, not a PR concern.
- **P3 (Errors typed, not thrown).** `Deferred.fail` with `RpcCallError` channel preserved.
- **P4 (Exhaustiveness).** Not touched.

### Non-blocking notes (out-of-scope per Amendment 4)

- The wire frame's `message` is still dropped on the registered-class path (the static class `message` becomes the instance message). The issue body scopes the fix to `data` only; `message` round-trip is a separate concern and may be filed independently if needed.
- No negative test for `data === undefined` ⇒ instance `data === undefined`. Preserved behavior is structural (the cast threads `undefined` straight through `Data.TaggedError`); the positive test plus the static type covers it. Worth adding a one-liner if a future regression is suspected, but not gating.

GitHub blocks self-approval on author-as-reviewer, so this comment carries the unconditional APPROVE verdict for the orchestrator's convergence gate.

Process issues: `gh pr view ... --jq '.conclusion // .status'` returns an empty string for in-progress checks (empty conclusion is truthy for `//`), so the suggested status-poll loop in the dispatch prompt would skip past pending CI; reworked to `select(.status == "IN_PROGRESS" or ...)` and confirmed CI before approving.

chughtapan:
## Verdict
**SHIP**

## Commands run
| Command | Exit | Log |
|---|---|---|
| `pnpm install` | 0 | /tmp/safer-verify-517/install.log |
| `pnpm build` | 0 | /tmp/safer-verify-517/build.log |
| `pnpm --filter=@moltzap/protocol test` | 0 | /tmp/safer-verify-517/test.log |

Head SHA: `b28b4e4f2e7b234452264f422616a4acd201b748`

## Test summary
Total: 115 (`@moltzap/protocol`) | Passed: 114 | Failed: 0 | Skipped/Todo: 1 | Flaky: 0

New test `src/transport/json-rpc-client.test.ts` (1 test) passes — registered tagged-error decode preserves wire `data` payload.

## Per-criterion check (against #511 acceptance)
- [x] **`json-rpc-client.ts:resolve` passes wire frame's `data` to registered error class constructor** — evidence: `packages/protocol/src/transport/json-rpc-client.ts:91-96@b28b4e4` constructs `new cls({ data: frame.error.data } as never)` rather than `new cls()`.
- [x] **Unit test asserts data round-trips** — evidence: `packages/protocol/src/transport/json-rpc-client.test.ts:74-77@b28b4e4` asserts `expect((value as ForbiddenError).data).toEqual(wireData)` after feeding a `ForbiddenError`-coded inbound frame carrying `{ agentIds: ["agent-a", "agent-b"] }`. Test passes locally and in CI.
- [x] **CI green** — `ci`, `test`, and `conformance` all SUCCESS on head `b28b4e4` per `gh pr view 517 --json statusCheckRollup`.

## Findings
None.

## Routing
Merge ready. Team-lead auto-merges per epic #512 contract.

PR #590, opened 2026-05-11
feat(client): per-conversation messageId dedup for inbound messages (#580)

### LINKED ISSUE #580: client: messageId dedup on messages/list and live broadcast (precondition for operator-driven re-broadcast)
## Why

Filed as a precondition for the deferred operator-driven recovery feature cited by [#463 v2](https://github.com/chughtapan/moltzap/issues/463#issuecomment-4414480360) §5.

#463 v1 R6 mitigation claimed "the recipient client's `messages/list` poll (or the live notification stream) deduplicates by `messageId`." Verified at HEAD: no such dedup logic exists. `git grep -nE 'messageId.*dedup|seen.*messageId|deduplicate' packages/client/src` returns no matches; `packages/client/src/service.ts:875` has a "seen cursor" for the cross-conversation peek API but not for live notifications.

## Residual risk #463 v2 closes by design

At HEAD without operator-driven recovery, the duplicate-broadcast risk does NOT occur: `sendCommit` calls `broadcast` exactly once per row, and the CAS in `recordBroadcastAttempt` only gates trace + delivery-webhook (which DOES make those at-most-once). #463 v2 R6 leans on that bounded-by-design property.

## Residual risk #463 v2 acknowledges remains

When operator-driven recovery ships (filed as deferred in #463 v2 §5), the broadcast itself becomes potentially-at-most-twice-per-row (first attempt + operator-triggered retry). At that point, the recipient client SHOULD dedup duplicate live broadcasts by `messageId`. That is the work this issue tracks.

## Acceptance

- Client-side seen-set keyed by `messageId` for live broadcast notifications.
- `messages/list` poll output deduped against the seen-set.
- Conformance test: server emits two broadcasts for the same `messageId`; client surfaces one.

## Out of scope

- The operator-driven recovery feature itself (deferred per #463 §5).
- Server-side de-emission (CAS guards trace + webhook, not broadcast emission).

## Trace

- #463 v2 §9 R6 cites this issue as the deferred mitigation for the duplicate-broadcast risk that surfaces only when operator-driven recovery is built.

chughtapan:
## Summary

Adds per-conversation LRU dedup of inbound messageIds on the **live `messages/received` channel**. Defensive scaffolding for any future retry mechanism (per arch-463 v3 — no automatic retry at HEAD means no current duplicate-broadcast risk, but lands future-proofing).

## Implementation

- `seenMessageIds: Map<ConversationId, Set<MessageId>>` on the service
- On inbound `messages/received`: check if messageId already seen → drop (no downstream dispatch)
- On miss: add to set + LRU evict if cap exceeded; proceed
- On conversation close: clear the set

Memory bound: ~1000 messageIds × 36 bytes UUID = ~36 KB per conversation. Bounded.

## Scope (Shape A — live-only dedup)

Re-review-590 (https://github.com/chughtapan/moltzap/pull/590#issuecomment-4424465901) caught a regression in an earlier attempt to extend the dedup window to the `messages/list` poll return. Filtering history through the live seen-set drops messages the consumer asked for, breaking 6 integration tests in `service.integration.test.ts`. Reverted in `62b53cd`.

**AC #2 reframing.** "dedup on poll" is unified as: the **live channel** dedups its own broadcasts; `messages/list` returns raw history unfiltered. Consumers that combine live + poll feeds dedup themselves (the seen-set is private to the live path; opening a public helper is deferred until a consumer requests it).

Architectural alternative (Shape B — unify live + poll into a single event emitter so dedup at the emitter ensures one `on("message")` event per id regardless of arrival route) is a larger change; revisit if there's demand.

## Tests

**Unit (`service.test.ts`, 217 tests):**
- Duplicate messageId on live channel → second invocation dropped
- Distinct messageIds → both processed normally
- LRU eviction → 1001st message kicks out 1st; 1st re-arriving is processed
- Conversation archive → dedup set cleared
- `close()` clears dedup window

**Conformance (`__tests__/conformance/dedup.test.ts`, 2 tests):**
- In-process `@effect/platform` WS server double-emits the same `MessageReceivedNotification` frame; `on("message", ...)` surfaces exactly one message
- Fresh id after a duplicate still surfaces (dedup window does not poison unrelated ids)

## Verification

- `pnpm --filter @moltzap/client build` green
- `pnpm --filter @moltzap/client test` — 217 pass + 6 todo
- `pnpm --filter @moltzap/client test:integration` — 40/40 pass (the 6 previously-failing tests pass)
- `pnpm --filter @moltzap/client test:conformance` — 12/12 pass
- `pnpm --filter @moltzap/protocol build` + `test` green
- Pre-commit hook clean without `--no-verify`

Fixes #580.
Refs arch-463 v3.

### DIFF
--- a/packages/client/src/__tests__/conformance/dedup.test.ts
+++ b/packages/client/src/__tests__/conformance/dedup.test.ts
@@ -1,0 +1,250 @@
+/**
+ * Conformance test for inbound messageId dedup: the server double-emits
+ * the same `messageId` over a real WS transport, and the client's
+ * `on("message", ...)` handler surfaces it exactly once.
+ *
+ * Exercises the full live-broadcast stack:
+ *   - in-process `@effect/platform` WS server
+ *   - `MoltZapService` connected through `MoltZapWsClient` (no fake)
+ *   - the service-level `on("message", ...)` handler is the
+ *     observation surface, downstream of the per-conversation
+ *     `seenMessageIds` window
+ *
+ * The dedup window lives at the `MoltZapService` layer, not the WS
+ * frame layer; this property anchors to the surface where the
+ * invariant actually binds (downstream handlers must surface ONE
+ * message, not two).
+ */
+import { describe, expect, it } from "vitest";
+import { Effect, Exit, Scope } from "effect";
+import * as NodeSocketServer from "@effect/platform-node/NodeSocketServer";
+import {
+  Connect,
+  MessageReceivedNotificationDefinition,
+  PROTOCOL_VERSION,
+  type RequestFrame,
+} from "@moltzap/protocol";
+import {
+  agentId,
+  conversationId,
+  messageId,
+  validateRequestFrame,
+} from "@moltzap/protocol/testing";
+import { MoltZapService } from "../../service.js";
+import type { Message } from "@moltzap/protocol";
+
+const LOCALHOST_HOST = "127.0.0.1";
+
+const TEST_AGENT_ID = agentId("11111111-1111-4111-8111-111111111111");
+const SENDER_AGENT_ID = agentId("22222222-2222-4222-8222-222222222222");
+const TEST_CONV_ID = conversationId("33333333-3333-4333-8333-333333333333");
+const TEST_MSG_ID = messageId("44444444-4444-4444-8444-444444444444");
+
+const TEST_POLICY = {
+  maxMessageBytes: 1_000_000,
+  maxPartsPerMessage: 10,
+  maxTextLength: 32_768,
+  maxGroupParticipants: 100,
+  heartbeatIntervalMs: 30_000,
+  rateLimits: {
+    messagesPerMinute: 60,
+    requestsPerMinute: 120,
+  },
+};
+
+const helloOk = () => ({
+  protocolVersion: PROTOCOL_VERSION,
+  agentId: TEST_AGENT_ID,
+  policy: TEST_POLICY,
+});
+
+const messageReceivedFrame = () =>
+  MessageReceivedNotificationDefinition.encode({
+    message: {
+      id: TEST_MSG_ID,
+      conversationId: TEST_CONV_ID,
+      senderId: SENDER_AGENT_ID,
+      parts: [{ type: "text", text: "double-emit-probe" }],
+      createdAt: "2026-05-11T00:00:00.000Z",
+    },
+  });
+
+interface ServerConn {
+  readonly send: (raw: string) => Effect.Effect<void>;
+}
+
+/**
+ * Spin up an in-process WS server that auto-responds to `network/connect`
+ * with `helloOk()`. Returns the bound URL and the most recently accepted
+ * server-side connection (used by the test to script post-handshake
+ * emissions).
+ */
+function startServer(): Effect.Effect<
+  { url: string; firstConn: Promise<ServerConn> },
+  unknown,
+  Scope.Scope
+> {
+  return Effect.gen(function* () {
+    const server = yield* NodeSocketServer.makeWebSocket({
+      port: 0,
+      host: LOCALHOST_HOST,
+    });
+    const addr = server.address;
+    if (addr._tag !== "TcpAddress") {
+      return yield* Effect.die("expected TcpAddress");
+    }
+    let resolveFirst!: (conn: ServerConn) => void;
+    const firstConn = new Promise<ServerConn>((res) => {
+      resolveFirst = res;
+    });
+    yield* Effect.forkScoped(
+      server
+        .run((serverSock) =>
+          Effect.gen(function* () {
+            const write = yield* serverSock.writer;
+            const conn: ServerConn = {
+              send: (raw) => write(raw).pipe(Effect.ignore),
+            };
+            resolveFirst(conn);
+            yield* serverSock.runRaw((data) =>
+              Effect.gen(function* () {
+                const raw =
+                  typeof data === "string"
+                    ? data
+                    : new TextDecoder("utf-8").decode(data);
+                const parsed: unknown = JSON.parse(raw);
+                if (!validateRequestFrame(parsed)) {
+                  // Drop non-request frames; the service only sends typed RPCs.
+                  return;
+                }
+                const frame: RequestFrame = parsed;
+                if (frame.method === Connect.name) {
+                  yield* write(
+                    JSON.stringify(Connect.encodeResponse(frame.id, helloOk())),
+                  ).pipe(Effect.ignore);
+                }
+                // Other RPCs are not exercised; the test only needs the
+                // post-handshake notification channel.
+              }),
+            );
+          }),
+        )
+        .pipe(Effect.ignore),
+    );
+    return {
+      url: `http://${LOCALHOST_HOST}:${addr.port}`,
+      firstConn,
+    };
+  });
+}
+
+async function withScope<A>(
+  effect: Effect.Effect<A, unknown, Scope.Scope>,
+): Promise<A> {
+  const scope = Effect.runSync(Scope.make());
+  try {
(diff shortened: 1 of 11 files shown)


chughtapan:
## safer:review-senior verdict: REQUEST-CHANGES

**Fallback reviewer** (no composed gstack skills in environment). Review per `safer:review-senior` SKILL.md fallback workflow. Head SHA: `83de0dd`.

---

## Scope gap — 2 of 3 acceptance criteria unmet

Sub-issue #580 names three acceptance criteria:

1. Client-side seen-set keyed by `messageId` for live broadcast notifications. **Met.**
2. `messages/list` poll output deduped against the seen-set. **Not met.**
3. Conformance test: server emits two broadcasts for the same `messageId`; client surfaces one. **Not met.**

**Item 2.** The `MessagesList` RPC handler (`packages/client/src/service.ts:623@83de0dd`) calls `sendRpc(MessagesList, ...)` and returns results directly. The `seenMessageIds` set is not consulted. Messages returned from `messages/list` can include IDs already in the seen-set with no dedup applied. The AC is unambiguous: poll output must be filtered through the same set.

**Item 3.** Zero conformance test files changed in this PR. The AC specifies a protocol-level conformance test (server-side double-emit, client surfaces one). The existing unit tests exercise the client's live-notification path via `FakeMoltZapService.emitEvent()` only, not the server-to-client transport contract. A conformance test at `packages/protocol/src/testing/conformance/` is required.

---

## P1-P4 craft notes

**P1 (Types beat tests) — nit.** `seenMessageIds` is typed `Map<string, Set<string>>` (`service.ts:306@83de0dd`). `ConversationId` and `MessageId` are branded types exported from `@moltzap/protocol`. The field should be `Map<ConversationId, Set<MessageId>>`. The weaker type silently accepts any `string` as a key, hiding confuse-a-ConversationId-with-a-MessageId errors from the compiler. Fix: one import + two type annotations.

**P3 — no finding.** The `as string` cast at the eviction site (`service.ts:1241@83de0dd`) is justified by the `size >= DEDUP_WINDOW_PER_CONV` guard immediately above.

**P2, P4 — no findings.**

---

## Test predicate audit

All six unit tests assert observable handler-invocation counts via `seen.length`. Each assertion fails if dedup is removed. LRU eviction test (`evict-msg-1` re-processed after 1001 fills) and archive/close tests correctly assert re-delivery after state is cleared. Not vacuous.

---

## CI status

`ci` job: green (1m38s). `test` and `conformance`: pending at review time. Verdict capped at APPROVE-PENDING-CI per Phase 1a gate, but scope gap overrides to REQUEST-CHANGES regardless.

---

## Pre-commit hook

No `--no-verify` or hook-skip in commit history.

---

## Required before merge

1. Dedup `messages/list` poll results against `seenMessageIds` (AC item 2).
2. Add conformance test: server double-emits same `messageId`, client surfaces one (AC item 3).
3. (Non-blocking, fix in same pass): change `seenMessageIds` type to `Map<ConversationId, Set<MessageId>>`.

**Verdict: REQUEST-CHANGES. Ship: NO-SHIP.**
Status: DONE_WITH_CONCERNS (fallback reviewer; CI `test`+`conformance` pending at verdict time).
Confidence: HIGH on scope gap (MessagesList handler confirmed not consulting `seenMessageIds`).
Process issues: gstack `/review`+`/simplify`+`/codex` unavailable; fallback mode activated.

chughtapan:
## safer:review-senior verdict: REQUEST-CHANGES

**Fallback reviewer** (no composed gstack skills in environment). Head SHA: `fe29a3a`.

---

## CI status: FAILING — 6 integration test regressions

CI gate per Phase 1a: the `test` job fails at `packages/client/src/__tests__/service.integration.test.ts` with 6 failing tests. These are **new regressions introduced by commit `fe29a3a`**, not pre-existing failures.

Failing tests (all in "Socket Server" suite):
- `history via socket returns messages with isOwn labels`
- `messages stay *NEW* after getContext notification until history is read`
- `new messages after history read are marked *NEW*`
- `different sessions have independent read markers`
- `lastRead tracks seen message IDs across reads`
- `non-text message parts render as markers in socket history`

**Root cause.** The poll-path dedup at `packages/client/src/service.ts:637-639@fe29a3a` applies `recordMessageIdIfNew` to every message returned by `messages/list`. In integration tests, the real server delivers a `messages/received` notification first (live broadcast), which records those message IDs in `seenMessageIds`. When the test subsequently calls `socketRequest("history", ...)`, the dedup filter drops those same IDs as duplicates, returning fewer messages than expected. Representative failure: `expected 1 to be >= 2` (line 1172); `expected 0 to be 1` (lines 1238, 1293, 1371, 1485).

The design intent — suppressing a message that arrived via live broadcast when the server also returns it in a poll — is architecturally sound for streaming dedup. But `messages/list` is a **history query**: callers expect all messages in the window regardless of what the live channel delivered. The AC says "dedup poll output against the seen-set" to prevent a message from being surfaced *twice* to the application, not to suppress it from history reads.

**Fix direction.** `recordMessageIdIfNew` on the poll path must not suppress messages the client already received live from appearing in history results. Two approaches:
1. **Read-only check on the poll path**: in the `messages/list` filter, call `this.seenMessageIds.get(convId)?.has(m.id)` to detect IDs already seen live (and skip re-emitting them as events), but still return the full result to the caller. Or, only dedup poll-vs-poll (same-call within-response duplicates) without crossing the live window.
2. **Record-on-poll, skip-on-live**: record poll results in the seen-set (so the live channel can skip re-surfacing them), but don't filter the poll results themselves — the history query should return everything.

Approach 2 likely matches the AC intent most closely: "dedup" means the app-level handler fires once, not that the history API omits the message.

---

## Fix 1 (Poll-path dedup): PARTIALLY MET — regression present

The code at `packages/client/src/service.ts:637-639@fe29a3a` is present and shared (DRY). The shared helper is correct. The application logic is wrong for the history-query use case (see CI failures above).

## Fix 2 (Branded types): MET

`seenMessageIds: Map<ConversationId, Set<MessageId>>` at `packages/client/src/service.ts:309@fe29a3a`. Import at line 45. The eviction cast at line 1262 (`as MessageId`) is justified by the `size >= DEDUP_WINDOW_PER_CONV` guard. P1 satisfied.

## Fix 3 (Conformance test): MET — architectural placement correct

`packages/client/src/__tests__/conformance/dedup.test.ts@fe29a3a` exercises the full transport stack: in-process `@effect/platform` WS server, real `MoltZapService` connected through `MoltZapWsClient` (no fakes), observation at the `on("message", ...)` handler downstream of `seenMessageIds`. Two tests: double-emit suppression and fresh-ID pass-through after duplicate. The reasoning that dedup lives at the service layer (not the wire-frame layer) is correct; a wire-level test would observe frame delivery, not the dedup invariant. Conformance job: PASS.

## P1-P4 craft notes

**P1 — pre-existing nit (not introduced by this PR).** `archivedConversationIds` at `service.ts:299@fe29a3a` remains `Set<string>` while `seenMessageIds` now uses `Set<MessageId>`. Not introduced here; not a blocking finding for this diff.

**P2, P3, P4 — no findings** on the new code.

---

## Required before merge

1. Fix the poll-path dedup so `messages/list` history queries are not filtered through the live-broadcast `seenMessageIds` window in a way that suppresses expected history results. The 6 failing integration tests are the acceptance bar — they must all pass.
2. Re-run CI green on all three jobs (`ci`, `test`, `conformance`).

The unit tests (service.test.ts poll-dedup suite) and conformance tests pass. Only the integration tests regress. The fix is scoped to `service.ts:637-639@fe29a3a`.

**Effort estimate:** (human: ~1 hour / CC: ~5 min)
**Confidence:** HIGH — failure mode directly evidenced by 6 integration test assertions mapping to the dedup filter application.
**Process issues:** CI `test` and `conformance` were IN_PROGRESS at review start; waited ~7 minutes for results. Phase 1a gate correctly blocked APPROVE until completion.

Status: DONE

Issue #696, opened 2026-05-24
[safer:architect] Track A — cursor-paginate the list RPC surface


chughtapan:
Parent: #693
Modality: architect
Depends on: #694 (spec, plan-approved)
Spec: https://github.com/chughtapan/moltzap/issues/693#issuecomment-4527246718 (Decision 1, Invariants 1-3; folds in #692)

Acceptance: an architect design doc (modules, interfaces, query plan, stubs) + N=2 plan review, for Track A ONLY:
- Apply the cursor-pagination convention (Decision 1): `{ limit?, cursor? } → { <collection>, nextCursor? }`, opaque `(sortKey, id)` cursor + `limit+1`, per Invariant 3.
- `TaskList`: wire `nextCursor` (thread `cursor` through `TaskService.list`, return `nextCursor`). **Track A owns ONLY the pagination ENVELOPE on `TaskList`; the `tasks` collection's ITEM type is owned by Track B and reshaped post-A. Do NOT touch the item content.**
- `ContactsList`: add `{ limit?, cursor? }` → `{ contacts, nextCursor? }`.
- `AgentsList`: convert `agents: Record<AgentId, AgentCard>` → `Array<AgentCard>` + cursor; enumerate + migrate every `AgentsList` consumer (use LSP/grep find-references).
- `MessagesList`: leave as-is (record the explicit decision).
- No raw SQL (Kysely); validate at boundary (TypeBox); no back-compat shims; conformance/canary/doc updates named.

**Track A LANDS FIRST; Track B rebases onto A's first CI-green merge.** Coordinate with Track B (#TBD-B) on the shared `TaskList` descriptor — A owns the envelope, B owns the item.

Use LSP (definitions/references/callers) to scope the real blast radius; cite by symbol name.

## Status
`planning`

chughtapan:
## [safer:architect] Track A design — cursor-paginate the list-RPC surface

Spec: https://github.com/chughtapan/moltzap/issues/693#issuecomment-4527246718 (Decision 1, Invariants 1-3) · Sub-issue: #696 · Parent epic: #693
Doctrine: safer-by-default v0.2.0 (`PRINCIPLES.md`, SHA on the #693 contract). Code citations pinned to `b683a93`.
Implementation tier: `implement-senior` (touches protocol + server + client CLI + conformance + docs; no new module, no new dep).

---

### 1. Summary

[...]

Issue #693, opened 2026-05-24
[safer:parent] Unify TaskList + cursor-paginate list surface + task-level DM dedup


chughtapan:
## Intent

Correct and consolidate the task/conversation listing + pagination surface. Triggered by codex's P2 on PR #691 (#685): client "one DM per pair" dedup matched *conversation* participants, which is wrong — it must match the *task* participant set, and the client has no task-participant data on the wire today.

## Context

- **Project:** moltzap (https://github.com/chughtapan/moltzap)
- **Motivation:** The D3 cutover (#669, merged) retired server-side DEFAULT_APP_ID dedup (#677) and moved "one DM per pair" to the client. The interim client dedup (PR #691) matches conversation participants, which incorrectly reuses a `{caller,bob}` sub-conversation inside a `{caller,bob,carol}` group task. Correct dedup is task-level; the wire exposes no task-participant set. A pagination audit (done while scoping this) also found `TaskList` half-wired (cursor in, no `nextCursor` out) and `ContactsList`/`AgentsList` unbounded.
- **Prior artifacts:**
  - https://github.com/chughtapan/moltzap/pull/691 — interim conversation-level dedup (to be superseded)
  - https://github.com/chughtapan/moltzap/issues/685 — client DM dedup (dead-code + proactive dedup)
  - https://github.com/chughtapan/moltzap/issues/692 — pagination audit / cursor-paginate the list surface
  - https://github.com/chughtapan/moltzap/issues/684 — stale-waiting-task reaper (unrelated follow-up)

## Contract

**Mode.** `refactor` — protocol-surface consolidation + correctness fix; findings addressed inline, not deferred; stamina at the high end.

**Goal.** Reshape `TaskList` to carry the task participant set + the caller's conversations and subsume `TaskConversationList` (then retire it); make the list-RPC surface consistently cursor-paginated; implement correct task-level "one DM per pair" dedup. Two implementation tracks — **A: pagination for the list surface**, **B: `TaskList` unification + dedup** — share the `TaskList` shape and merge in a coordinated order.

**Acceptance.**
- [ ] `/safer:contract` spec published defining: cursor-pagination convention; unified `TaskList` item shape (task participants + caller's conversations); dedup match rule + conversation-selection rule (DM task has one conversation; >1 ⇒ not a DM, don't dedup); `TaskConversationList` retirement + consumer migration (`service.ts`×2, `openclaw`, `start.ts`); A/B scope split incl. `TaskList` merge order.
[...]

Issue #694, opened 2026-05-24
[safer:contract] Spec: unified TaskList + pagination convention + task-level DM dedup


chughtapan:
Parent: #693
Modality: contract
Depends on: —

Acceptance — a spec published (comment on parent epic #693, or this sub-issue body) defining ALL of:
1. **Cursor-pagination convention** — one shape (mirror `TaskConversationList`: `limit` + optional `cursor` → `{ items, nextCursor? }`); name which list RPCs adopt it.
2. **Unified `TaskList` item shape** — `Task` fields + the TASK participant set + the caller's conversations (caller-scoped; archived included w/ `archivedAt`). Decide whether per-conversation participants stay on the wire (no current consumer reads them).
3. **Dedup match rule + conversation-selection rule** — match the TASK participant set `== {caller} ∪ invitedAgentIds` (caller implicit ⇒ `invited ⊆ participants` ∧ `|participants| == |unique(invited)|+1`). A DM task has ONE conversation; >1 ⇒ not a DM ⇒ do not dedup (NOT "first non-archived").
4. **`TaskConversationList` retirement + consumer migration** — consumers: `client/src/service.ts` (`fetchHistoryConversationMeta`, `resolveReusableConversationId`), `openclaw-channel/src/openclaw-entry.ts` (`listGroupsEffect`), `client/src/cli/commands/start.ts` (dedup). All use `{taskId, conversation}` (+ pagination in 2); none read conversation participants.
5. **A/B scope split + `TaskList` merge order** — Track A (pagination: ContactsList, AgentsList Record→array, TaskList nextCursor, MessagesList decision) vs Track B (TaskList content reshape + dedup + retire TaskConversationList). Define ownership of `TaskList`'s final shape and the land/rebase order.

Pagination audit (input): `TaskList` accepts `cursor` but returns no `nextCursor`; `ContactsList`/`AgentsList` unbounded (`AgentsList` is a `Record`); `MessagesList` uses `sinceSeq`+`hasMore`; `TaskConversationList` is the working cursor model.

Constraints: no raw SQL (Kysely); no back-compat shims (migrate in-PR); validate at boundaries; cite by symbol name. Use LSP to verify consumers + blast radius.

## Status
\`planning\`

chughtapan:
## N=2 plan review — spec → REVISE (both passes)

[...]

Issue #673, opened 2026-05-21
Refactor: delete tm_endpoint_address column + EndpointAddress brand; route TM by tasks.app_id


chughtapan:
## Problem

`tasks.tm_endpoint_address` is pure redundancy with `tasks.app_id`. The address is always `"tm:app:" + app_id`, written via `defaultAppTmEndpointAddress(appId)` and read via `appIdFromTmEndpointAddress(addr)` — a serialize/parse round-trip that adds zero information.

The duplication actively hurts the type system. The column is typed `string` and accepts both `"tm:app:<id>"` and `"tm:agent:<id>"`. Pre-D3 had two producers (`TasksCreate({tmType: "self"})` wrote agent variants; default-app-TM wrote app variants). D3 deleted the agent-producing path on the wire but left `endpointAddressForAgent` in production, the brand intact, and the read-side check (`loadTaskAsTmAuthority`) comparing against `tm:agent:<caller>` — which never matches in prod and is therefore always-false.

## Proposed deletion

- Drop `tasks.tm_endpoint_address` column.
- Delete `EndpointAddress` brand from `@moltzap/protocol/network`.
- Delete `endpointAddressForAgent`, `defaultAppTmEndpointAddress`, `appIdFromTmEndpointAddress`, `brandEndpointAddress`, `makeEndpointAddress`.
- `loadTaskAsTmAuthority(taskId, callerConnId)` → look up `tasks.app_id` → check `appHost.remoteRegistrations.get(appId)?.connectionId === callerConnId` (or `hooks.has(appId)` for in-process apps).
- `network-send.ts` dispatch paths route by `appId` directly.
- `app-host.ts → messageAuthorize` / `dispatchAuthorize` take `appId` instead of `tmEndpointAddress`.
- `message-send-permission.ts` sender-is-TM check becomes \"caller's connId == remoteRegistrations[task.appId]?.connectionId.\"

## What this fixes

- The wire `TaskConversation*` / `TaskClose` / `TaskAddParticipant` etc. RPCs become reachable by `MoltZapTMClient` callers that have `AppsRegister`'d their app. Today they're structurally unreachable because the byte-equality check requires `tm:agent:<caller>` which no producer ever writes.
- Eliminates the foot-gun by construction: the agent variant of the address simply doesn't exist as a type. The compiler enforces \"TM is an app\" instead of relying on a runtime byte comparison that's silently always-false.
[...]

PR #700, opened 2026-05-24
[impl-senior] Track A — cursor-paginate list surface (#692)

### LINKED ISSUE #692: Cursor-paginate the list RPC surface (TaskList nextCursor, ContactsList, AgentsList)
## Context

Pagination audit (done while scoping #685) found the list-returning RPC surface is inconsistent — one half-wired, two fully unbounded. This issue makes cursor pagination consistent across the surface.

| RPC | Result | Pagination | State |
|---|---|---|---|
| `TaskList` | `{ tasks: Task[] }` | `limit`≤200 + `cursor` **in**, no `nextCursor` **out** | **BROKEN** — accepts a cursor it never returns |
| `ContactsList` | `{ contacts: Contact[] }` | none | **unbounded** |
| `AgentsList` | `{ agents: Record<AgentId, AgentCard> }` | none | **unbounded** + map-shaped (not array) |
| `MessagesList` | `{ messages[], hasMore }` | `sinceSeq` + `limit` → `hasMore` | functional but asymmetric (no `nextCursor`) |
| `TaskConversationList` | `{ items[], nextCursor? }` | `limit` + `cursor` → `nextCursor` | ✅ reference model |
| `AgentsLookup`/`AgentsLookupByName`/`PresenceSubscribe` | request-bounded | n/a | ✅ |

## Goals

- Establish ONE cursor-pagination convention (mirror `TaskConversationList`: `limit` + `cursor` → `{ items, nextCursor? }`).
- Fix `TaskList` to return `nextCursor` (close the half-wired pagination).
- Add cursor pagination to `ContactsList`.
- Reshape `AgentsList` from `Record<AgentId, AgentCard>` to a cursor-paginated array (wire break — migrate consumers).
- Decide whether to standardize `MessagesList` (`hasMore` → `nextCursor`) or leave it (it works).

## Non-goals

- The `TaskList` CONTENT reshape (task participants + caller's conversations) and `TaskConversationList` retirement — that is the sibling issue (Task* unification). **Overlap:** both touch `TaskList`. This issue owns its *pagination mechanics*; the sibling owns its *payload shape*. Coordinate at merge (likely: land one, rebase the other onto the final `TaskList` shape).
- Backend storage/index changes beyond what cursor pagination needs.

## Invariants

- Cursor is opaque to clients (server-defined encoding; today an activity/created-at anchor).
- A page miss/empty result is not an error.
- Pagination stays caller-scoped (no cross-agent leakage).
- No raw SQL — Kysely query builder only.

## Acceptance criteria

- `TaskList`, `ContactsList`, `AgentsList` accept `limit` + optional `cursor` and return `nextCursor?`.
- `AgentsList` consumers migrated off the `Record` shape.
- Conformance + canary + integration tests updated; idempotence-across-replay holds.
- Docs/ARCHITECTURE updated.

## Open questions (for the architect)

1. `AgentsList` `Record → array` is a wire break — who consumes it (LSP: find references) and what is the migration?
2. Standardize `MessagesList` or leave it?
3. Shared `Paginated<T>` schema helper, or per-method shapes?

Use LSP (definitions/references/callers/implementations) to scope the real blast radius, not grep windows. Refs #685.

chughtapan:
Closes #692
Architect plan: https://github.com/chughtapan/moltzap/issues/696#issuecomment-4527715414
Spec: https://github.com/chughtapan/moltzap/issues/693#issuecomment-4527246718 (Decision 1, Invariants 1-3)

## What changed
Applies the spec's ONE cursor-pagination convention to the list-RPC surface: `{ limit?, cursor? } → { <collection>, nextCursor? }`, opaque `(created_at, id)` keyset cursor, `limit + 1` overflow probe. Track A owns only the pagination envelope; Track B reshapes the `TaskList` item later.

## Plan anchors
| Change | Plan anchor |
|---|---|
| `AgentsList` Record→Array + cursor envelope | §3.2, §8 |
| `ContactsList` cursor envelope (additive) | §3.3, §9 |
| `TaskList` add `nextCursor` + brand cursor (envelope only; item stays `Task[]`) | §3.4, §12 |
| branded `ListCursor` in `schema-primitives.ts` (`listCursorSchema`, no uuid format) | §3.1 |
| new `packages/server/src/db/list-cursor.ts` codec (`encode`/`decode`/`keysetWhere`/`paginate` + `InvalidCursorError`) | §3.5, modules §2 row 8 |
| `TaskService.list` thread cursor, `limit+1`, return `{ tasks, nextCursor? }` | §3.6 |
| `ContactsService.list({ limit?, cursor? })` → `{ contacts, nextCursor? }` | §3.7 |
| `AgentsList` handler → `AgentCard[]` keyset-paginated | §3.8 |
| handlers map `InvalidCursorError` → `InvalidParamsError` | §5 |
| CLI `agents`/`contacts` list: Array shape + first-page marker | §8 row 4, §9, #696 approval |
| conformance `{agents:{}}`→`{agents:[]}`, new pagination canary, docs | §11 |
| `MessagesList` left as-is (recorded) | §10 |

## AgentsList migration (Record → Array)
6 sites migrated in-PR (no back-compat, no external consumers): descriptor (`identity/agents.ts`), server handler (`agents-lookup.handlers.ts`), CLI `agents list` (`Object.values`→array), integration test (`agents-list.test.ts` map-index→`find`), conformance emitter (`rpc-semantics.ts` `{agents:{}}`→`{agents:[]}`), divergence-proof bad-server shape. Re-exports verified no-op.

## CLI first-page marking (#696 approval decision)
`agents list` and `contacts list` print `Showing first N — more results available.` when the result carries `nextCursor`, so a truncated page is never mistaken for the full list. No auto-loop. (No `tasks` list CLI command exists; the `TaskList` client consumers are Track B's.)

## Notable fix found in review
`/review` codex pass caught a real keyset bug: `created_at` is `TIMESTAMPTZ` (microsecond) but the cursor `sortKey` is `Date.toISOString()` (millisecond), so the predicate could skip rows sharing a millisecond but differing in microseconds. Fixed by truncating `created_at` to milliseconds (`date_trunc`, builder-only, no raw SQL) in BOTH the order-by and the keyset predicate, so column resolution matches the cursor. New `task.service.test.ts` regression (`paginates sub-millisecond created_at ties without skips or dups`) fails against the pre-fix codec and passes after — verified non-vacuous.

Conformance also surfaced a model-equivalence divergence: `agents/list` left the K=1 confident-oracle set because a server-validated `cursor` makes a schema-valid-but-undecodable token yield `InvalidParamsError`. `MODEL_METHOD_OUTCOMES[AgentsList] "ok" → "uncertain"`, matching every other cursor-bearing list RPC.

## Scope
- Modules touched: protocol (identity, task, schema-primitives, index, testing), server (db/list-cursor [new], task.service, contact.service, agents-lookup handler, contacts/tasks handlers, kysely-vendor, eslint guard), client CLI (agents, contacts), docs.
- New modules: 1 (`db/list-cursor.ts`) — plan-named (§2 row 8, §3.5). New public surface: `ListCursor` + `listCursorSchema` — plan-named (§3.1). New deps: 0.
- `safer-diff-scope`: reports `staff` (37-file count heuristic). Shape is senior: 0 new modules outside the plan, 0 new deps, 0 new public contracts outside the plan; the file count is driven by the plan-mandated AgentsList Record→Array consumer migration. Flagged for review-senior to adjudicate.

## Cursor opacity / tampering (deliberate non-finding)
The cursor is opaque, NOT authenticated (Invariant 2). A client can forge a valid-shaped cursor, but the keyset filter runs AFTER caller-scoping (`task_participants.agent_id = caller`, `owner_user_id = owner`, `visibleAgentIds`), so a forged cursor can only skip/reorder the caller's OWN results — no cross-agent leak (Invariant 4 holds). Matches the legacy `TaskConversationList` model.

## Simplify / Review
- `/simplify`: not available in this environment (no SKILL on disk) — noted; manual simplification pass applied (extracted `agentsListPage` helper, no dead code).
- `/review`: run on the diff; findings applied (the millisecond-precision keyset fix + the model-equivalence fix above). No remaining unresolved findings.
- `/document-release`: docs synced in-PR (method mdx via generator + brand description, ARCHITECTURE indexes, new `list-pagination.md` flow doc cross-linked, CHANGELOG entry).

## Tests
- protocol 141, server unit 196 (incl. sub-ms pagination regression), server conformance 45, client unit 299, client conformance 11, integration (agents-list 14 / contacts+tasks+agents 23), codec unit 6 — all green.
- build + typecheck + lint + format all pass.

## A/B coordination
`TaskList` descriptor: Track A touches ONLY the pagination envelope (`nextCursor` + branded `cursor`); the `tasks` item stays `Type.Array(TaskSchema)`. Track B reshapes the item to `TaskListItem` on rebase and owns the FINAL `TaskList` canaries (this PR ships an interim `nextCursor` canary flagged as such). Land A first, then B rebases (Decision 5).

## Confidence
HIGH — full gate suite green; the two cross-cutting interactions (keyset ms-precision, model-equivalence confidence) were caught by review + conformance and fixed with regression coverage.


chughtapan:
## Verdict: APPROVE

**review-senior round 1.** Reviewer: Claude (`/safer:review-senior`). The other N=2 pass (`/codex`) runs separately. Merge bar is 2 consecutive clean rounds per reviewer; this is round 1 clean from this reviewer. Merge stays USER-GATED.

CI status: green (`ci`, `conformance`, `test` all SUCCESS; Mintlify skipped).

Reviewed at SHA `81c5dec0`. Diff is the single-commit branch vs merge-base `95a3c49`; the PR's three-dot stat lists a phantom `server-executable.proofs.test.ts` entry that is a main-vs-branch divergence artifact (the file was moved on main; no PR edit touches it).

### Highest-risk targets — all verified

**1. Keyset cursor correctness (PASS, regression test proven non-vacuous).** `sortKeyExpr` applies `date_trunc('milliseconds', created_at)` and is passed to BOTH `.orderBy(...)` and `keysetWhere` in all three query sites (`task.service.ts:346,355`; `contact.service.ts:77,83`; `agents-lookup.handlers.ts:90,96`), so the page boundary and the predicate use the identical expression. Both comparison sides are ms-resolution (truncated `timestamptz` column vs the ms `Date.toISOString()` cursor), and the `(trunc(created_at), id)` pair with `id ASC` tie-break is a strict total order matching ORDER BY exactly — no skip, no dup at the ms boundary. I proved the regression test (`task.service.test.ts → paginatesSubMillisecondTiesWithoutSkips`) is NON-VACUOUS: temporarily reverting `sortKeyExpr` to the raw-column form fails the test 2/4 (it skips `aa01`/`aa02`, the same-ms-earlier-µs siblings); the shipped `date_trunc` form passes all 4. Test seeds 3 rows sharing `.500` ms (differing µs) + 1 in a later ms and walks every page with `limit:1`.

**2. Cursor opacity + boundary (PASS, guard proven non-vacuous).** `ListCursor` is a branded type whose only construction site is the codec (`list-cursor.ts → encodeListCursor`); callers cannot construct or inspect it. I verified the `cursorOpacityGuard` in `eslint.config.mjs` FIRES: a throwaway file calling `atob(token)` and `Buffer.from(token,"base64url")` outside the codec produced 2 lint errors. A schema-valid-but-undecodable cursor maps to `InvalidParamsError` (not 500, not silent empty) in all three handlers via `Effect.catchTag("InvalidCursor")` (`agents-lookup.handlers.ts:191`, `contacts.handlers.ts`, `tasks.handlers.ts`).

**3. Record→Array migration (PASS, complete).** Grep-verified no straggler treats `AgentsList` as a `Record`: every remaining `result.agents[N]` is array index access (numeric), not string-key access. Migrated consumers: descriptor (`agents.ts`), handler, CLI (`cli/commands/agents.ts`), integration test (`agents-list.test.ts` Record→Array + `Object.keys`→`.map`/`.find`), conformance emitters (`rpc-semantics.ts` `{agents:{}}`→`{agents:[]}`), and the `identity-list-pagination.types-check.ts` canary asserting `AgentsListResult["agents"] === AgentCard[]`.

**4. No raw SQL / no N+1 (PASS).** Pure Kysely builder composition; no `sql\`` templates in any changed file. `agents/list` runs 2 queries (visible-ids filter + paginated row fetch over the visible set) — no per-row lookup. `task/list` uses a single `innerJoin task_participants` filtered to `agent_id = caller` (PK `(task_id, agent_id)` → at most one join row per task, no fan-out dup).

**5. Conformance oracle (PASS, justified).** `MODEL_METHOD_OUTCOMES[AgentsList.name]` `"ok"→"uncertain"` is correct: `agents/list` now takes a server-validated cursor that can yield `InvalidParamsError`, so the outcome is no longer param-invariantly ok. Consistent with every sibling cursor-bearing list RPC (`contacts/list`, `task/list`, `messages/list`, `task/conversation/list`), all `uncertain`. `model-equivalence.ts` K-set correctly narrows to `network/ping` only. Divergence-proof gate passes.

### P1-P4 craft floor (reviewer-owned)
- P1 (types): branded `ListCursor`, single construction site. PASS.
- P2 (boundaries): `decodeListCursor` validates base64url + canonical ISO-8601 + UUID at the wire boundary, fails typed. PASS.
- P3 (typed errors): `InvalidCursorError` is a `Data.TaggedError`; service `list` carries it in the error channel; handlers `catchTag` to `InvalidParamsError`. The only `throw` is `absurdTaskStatus` (exhaustiveness guard). PASS.
- P4 (exhaustiveness): `loadOpenTask` status switch ends in `absurdTaskStatus`. PASS.

### Two flagged concerns — rulings

**`safer-diff-scope` reports `staff` (37/33-file heuristic): RULING — correctly senior-shaped.** Ran `safer-diff-scope --pr 700`: `{tier:"staff", files:33, new_deps:0, rationale:"33 files changed — large-surface diff"}`. The tier verdict is driven purely by file count, not by change shape. The shape is senior: 0 new package deps, the one new module (`db/list-cursor.ts`) is plan-mandated by spec #693 Decision 1, and the public-surface additions (`ListCursor` brand + envelope on three RPCs) are exactly what the approved plan names. The bulk of the 33 files is the plan-mandated `Record→Array` consumer migration — mechanical fan-out, not architectural reach. This is a senior change whose file count tripped a volume heuristic; per Budget Gate doctrine, shape governs, not volume. No staff escalation was warranted.

**`/simplify` unavailable, implementer ran a manual pass: RULING — adequately simplified.** `/simplify` SKILL is absent from this environment (partial-miss, noted below). Assessed simplification myself: the codec factors cleanly into `encodeListCursor`/`decodeListCursor`/`sortKeyExpr`/`keysetWhere`/`paginate` with no duplication; all three services/handlers call the shared helpers identically; the cursor→`InvalidParamsError` mapping is the same 4-line `catchTag` in each handler (a shared helper would add indirection for 4 lines × 3 sites — not worth it). No avoidable complexity. One cosmetic nit (non-blocking): `agents-list.test.ts` keeps a now-redundant `AgentsArrayResult` alias alongside the migrated `AgentsListResult`; harmless.

### Unavailability note
PR-kind routing requires `/review` + `/simplify` + `/codex`. `/simplify` SKILL is absent from this environment (partial miss). The reviewer-owned P1-P4 craft floor + full diff read substitute for the missing convergence signal; team-lead should restore `/simplify` for future runs. Verdict tagged `DONE_WITH_CONCERNS` for the missing skill; the substantive correctness verdict is APPROVE.

No P1 or P2 findings. P3: one cosmetic redundant type alias (above), non-blocking.

APPROVE.

chughtapan:
## Orchestrator — N=2 round 1 consolidation

**review-senior:** APPROVE (clean round 1). 0 P1 / 0 P2; one P3 cosmetic (redundant `AgentsArrayResult` alias in `agents-list.test.ts`). Empirically proved (a) the keyset regression test is non-vacuous (reverting `sortKeyExpr` to raw-column form fails it 2/4) and (b) the opacity lint guard fires. Ruled both flagged concerns in the implementer's favor: `safer-diff-scope=staff` is a pure file-count heuristic, shape is senior; `/simplify`-absent code is adequately simplified.

**codex (`exec review`, gpt-5.5/medium):** one **P2** — `contacts/list` default page-size (50) silently truncates for `{}` callers ignoring `nextCursor`.

**Adjudication (orchestrator):** codex's *external-compat* framing is moot (no external consumers; bounding the lists is spec #693 Decision 1). But the finding is a **real in-repo regression** for a different reason: `openclaw-channel/src/openclaw-entry.ts → listPeersEffect` enumerates `ContactsList {}` → `flatMap(contactAgentIds)` to build the peer directory for target resolution, and now silently drops peers past position 50. **VALID P2.**

**Fix (routed to the implementer):** consumer-side pagination — openclaw pages through `nextCursor` to assemble the full set; the RPC default stays bounded (no un-bounding shim). Plus a non-vacuous >50-contact regression test, a class-sweep of other `{}` enumerate-all consumers of the newly-paginated lists, and the P3 alias deletion.

**Round 1 = NOT clean** (one valid P2). Re-review both legs after the fix; the 2-consecutive-clean-per-reviewer count starts from the post-fix SHA. Merge remains user-gated.

chughtapan:
## Verdict: APPROVE

**review-senior round 2** (toward 2-consecutive-clean; round 1 APPROVE stands). Reviewer: Claude (`/safer:review-senior`). Reviewed at SHA `e75855be`. Merge stays USER-GATED.

CI status: green (`ci`, `conformance`, `test` all SUCCESS; Mintlify skipped).

Delta since round-1 SHA `81c5dec0` is 5 files / 278 insertions. Reviewed the delta hard and re-confirmed the whole (round-1 keyset regression + codec, 20 tests, still green at this SHA).

### Fix-roll delta — all verified

**1. `drainPaginatedList` (`openclaw-entry.ts`).** Termination: loop ends when a page omits `nextCursor`; against the real server this is guaranteed (keyset strictly advances over finite rows, `paginate` emits `nextCursor` only on overflow). Opaque echo: `cursor = nextCursor` then `params = { cursor }` — echoes the token unmodified, never parses it (Invariant 2). Typed errors: returns `Effect.Effect<ResultOf<D>[K], ServiceRpcError>`, no bare `Promise`. No value-level `any` (the `any` is constraint-position `RpcDefinition<string, any, any>`, standard for a generic RPC helper; `ParamsOf`/`ResultOf`/`RpcDefinition` are real protocol exports and the casts are localized generic bridges). Typecheck clean.

**2. `lookupAgentsInChunks` (boundary at exactly 100).** `slice(i, i+100)` stepping `i += 100`: chunks are [0,100), [100,130), ... — no drop, no dup, no overlap; empty input → `[]`. 130 ids → 100 + 30. Correct. `listPeersEffect` drains ALL contact pages then chunks the lookup, so a >100-peer directory no longer trips `agents/lookup`'s `maxItems:100` cap.

**3. New `openclaw-entry.directory.test.ts` — NON-VACUOUS, confirmed empirically.** Drives `plugin.directory.listPeers` against a fake `sendRpc` paginating 130 contacts in 50-row pages (3 pages). Asserts all 130 enumerate (boundary names peer-0/49/50/129) and exactly `ceil(130/50)=3` `contacts/list` calls fire. I temporarily reverted `listPeersEffect` to single-page consumption: both tests FAIL (1 call / 50 peers); the drain-all version passes both. The fake replaces only `sendRpc` (RPC transport) — it does NOT mock dispatch or delivery, so it complies with the openclaw rule (`CLAUDE.md`: "Never mock the dispatch or delivery mechanism"; directory enumeration is a read-path, dispatch/delivery is untouched). This is a unit test of the read path, allowed.

**4. CLI `--json` (`agents.ts` / `contacts.ts`).** `--json` now serializes the full result object `r` (collection + `nextCursor`) instead of just the array, so machine consumers see the truncation signal. The human-path first-page footer (`Showing first N — more results available.`) is unchanged. Correct.

**5. P3 fold-in.** Redundant `AgentsArrayResult` alias deleted; its two uses (`lookupAgents`, `lookupAgentsByName`) repoint to `AgentsListResult` (the extra optional `nextCursor?` is harmless on lookup results). Clean.

### TaskConversationList out-of-scope ruling — AGREE, correctly out of Track A scope
`listGroupsEffect` consumes `task/conversation/list` single-page and ignores its `nextCursor`. Verified this is pre-existing and out of scope:
- The PR's diff hunks do NOT touch `listGroupsEffect` (`merge-base..e75855be` on `openclaw-entry.ts` shows no `listGroupsEffect`/`TaskConversationList` change). `client/src/service.ts` is not touched by the PR at all.
- `task/conversation/list` is the LEGACY cursor surface (plain `Type.String()` cursor, `limit` cap 100, NOT the branded `ListCursor`) the round-1 architecture doc explicitly names as predating this convention: "It is the model the convention supersedes; do not extend it." It is Track B's surface (capped in #669).
- Fixing it here would expand scope past spec #693 Decision 1 (the three RPCs `agents/list` / `contacts/list` / `task/list`) AND extend the legacy cursor the convention says not to extend.

Reporting-not-fixing is the right call. Process note (not a blocker): file a GH issue for the latent single-page `listGroupsEffect` + `client/src/service.ts` consumption so Track B picks it up, rather than leaving it as a tick-comment flag.

### P1-P4 craft floor (reviewer-owned): clean
P1 branded `ListCursor` opacity preserved (token echoed as opaque `string`, never decoded in `openclaw-channel`). P2 boundary unchanged from round 1. P3 `drainPaginatedList`/`lookupAgentsInChunks` carry typed `ServiceRpcError` channels. P4 n/a (no new union switch).

### Findings
- P1: 0. P2: 0.
- **P3 (non-blocking): `drainPaginatedList` has no termination guard against a non-advancing server.** `while (more)` loops forever if a misbehaving server returns a constant non-null `nextCursor` (the surrounding `catchAll` cannot interrupt the loop). Not a correctness bug against the real server, and the blast radius is a degraded directory read. Note: the repo's sibling drain (`client/src/service.ts → SEND_TO_AGENT_LOOKUP_MAX_PAGES = 20`) bounds its loop — but a naive page-cap on a COMPLETE drain would silently truncate a legitimate large directory, so the right hardening here is cursor-progress / cycle detection (stop + typed-error if `nextCursor` repeats), not a fixed page cap. Recommend for a follow-up; does not block this PR.

No blocking findings. Clean round.

APPROVE.

chughtapan:
## Orchestrator — N=2 round 2 consolidation (fix-roll `e75855be`)

**review-senior:** APPROVE — **2nd consecutive clean → leg SATISFIED.** Empirically re-proved drain-all non-vacuity (reverted to single-page → 1 call/50 peers fail; drain → 3 calls/130 pass), verified the 100-id chunk boundary (130 → 100+30, no drop/dup), `--json` now carries `nextCursor`, `AgentsArrayResult` cleanly removed. 0 P1 / 0 P2.

**codex (gpt-5.5/medium):** one **P2** — `list-cursor.ts` carries a literal **NUL byte** in a doc comment; `git diff` reports the file as `Bin 0 -> 6748 bytes` (whole file binary → diff/review/text-lint blind to it). Confirmed at HEAD. → **not clean.**

**One P3 (review-senior, non-blocking):** `drainPaginatedList` `while(more)` has no guard against a byzantine non-advancing server → uninterruptible infinite loop. Correct fix = cursor-progress/cycle detection, NOT a page cap (a cap would truncate legit large directories).

**Actions:**
- **Fix-roll #2 routed** (in flight): remove the NUL (+ class-scan no source file carries one), and fold the P3 drain-guard (progress/cycle detection). codex would re-flag the P3 next round, so fixing-while-disclosed beats a route cycle.
- **Filed https://github.com/chughtapan/moltzap/issues/702** — the out-of-scope `TaskConversationList` single-page consumption (`listGroupsEffect` + `service.ts`), for Track B's retirement to absorb. review-senior + I both rule it correctly out of Track A scope.

**Bar status:** review-senior satisfied (2 consecutive clean). codex needs 2 consecutive clean — 0 so far; re-review after the fix-roll. Merge user-gated.

chughtapan:
## Fix-roll #2 landed — `8dfd5a44`

Completed by the orchestrator (the impl agent hit a Claude session limit mid-roll; its edits were verified + finished rather than lost).

- **codex P2 (NUL byte) — RESOLVED.** `list-cursor.ts` had a literal NUL in a doc comment (Write-tool artifact) making git treat the whole file as binary. Replaced with a space matching `CURSOR_FIELD_SEP`. Committed file verified `tr`-clean (0 NUL); diffs as text again.
- **review-senior P3 (drain infinite-loop) — RESOLVED.** `drainPaginatedList` now does cursor cycle-detection: fails typed (`MoltZapNonAdvancingCursorError`) on a non-advancing `nextCursor` rather than looping forever. NOT a page cap (which would truncate large directories). New byzantine-constant-cursor test asserts termination.

Gates: build ✓, lint ✓, openclaw unit **61 passed** (was 60; +1 byzantine test). CI running on the pushed SHA.

**Note:** branch is 23 behind `origin/main` (overlap = 3 doc/changelog files only — Track A's base is already post-#673). Rebase onto current main pending before the merge gate.

chughtapan:
## Verdict: APPROVE

**review-senior final-confirm round** on SHA `e41e5f0c` (rebased onto current `origin/main`, merge-base `f027c33`). Reviewer: Claude (`/safer:review-senior`). This is clean #1 on the final SHA (prior 2-clean was on `e75855be`, now superseded by the rebase + 3 fixes). Merge stays USER-GATED.

CI status: green (`ci`, `conformance`, `test` all SUCCESS; Mintlify skipped).

Reviewed the 3-commit delta since `e75855be` hard, then re-confirmed the whole post-rebase.

### Delta — all verified

**1. `drainPaginatedList` cycle guard (my prior P3, now fixed).** `seenCursors: Set<string>` records each `nextCursor`; a repeat fails typed `MoltZapNonAdvancingCursorError` instead of looping. NOT a page cap (the comment cites the truncation risk I flagged). Logic is sound: a correct server emits a strictly-new `(sortKey,id)` cursor per page (injective codec, unique PK tie-break, keyset strictly excludes emitted rows), so no false-positive on legit pagination; the last page omits `nextCursor` and ends the loop normally. The error joins the typed channel (`ServiceRpcError | MoltZapNonAdvancingCursorError`), and `listPeersEffect`'s `catchAll(() => succeed([]))` degrades to an empty directory rather than hanging. The new byzantine test is NON-VACUOUS, proven empirically: with the guard removed the test does not pass — the unbounded loop exhausts memory (`RangeError: Invalid array length` after ~10.5s); with the guard it terminates after exactly 2 `contacts/list` calls and resolves to `[]`.

**2. NUL byte removed from `list-cursor.ts`.** `tr -cd '\000' < packages/server/src/db/list-cursor.ts | wc -c` → 0. Git now treats the file as text (numstat 191/0, no longer `Bin`). The doc comment reads correctly with the `CURSOR_FIELD_SEP` space.

**3. base64url canonical-form fix in `decodeListCursor` (codex round-4 P2).** Decode now checks `buf.toString("base64url") === cursor` and fails `InvalidCursorError` on any non-canonical token. Correct: a token produced by `encodeListCursor` round-trips canonically (no false-reject — verified the roundtrip property test still passes against the fixed decoder); `encode(pos)+"!"` re-encodes WITHOUT the `!` so `!== cursor` → reject. The new test `rejects a valid payload with trailing non-base64url junk` is NON-VACUOUS, proven empirically: against the permissive decoder it FAILS (`expected undefined to be an instance of InvalidCursorError` — the permissive decoder accepted the junk-suffixed token); against the canonical-fix it passes. Boundary mapping intact: handler `catchTag("InvalidCursor") → InvalidParamsError`.

**4. Rebase onto current main (incl. #673/OTel/#680) — Track A intact.** Independently re-confirmed post-rebase: server-core keyset regression + codec + contact-service (34 tests) green incl. `paginatesSubMillisecondTiesWithoutSkips`; openclaw full suite (61 tests) green; protocol typecheck (the `identity-list-pagination.types-check.ts` canary) compiles; the cursor-opacity eslint guard still FIRES (re-probed: `atob` + `Buffer.from(...,"base64url")` outside the codec → 2 errors); divergence-proof gate OK (42 registrars); no Record stragglers reintroduced (`Record<string, AgentCard>` / `Object.keys(...agents)` → none). Builds clean (protocol, client, openclaw). Lint clean on all changed server + openclaw source incl. `eslint.config.mjs`.

### P1-P4 craft floor (reviewer-owned): clean
P1 branded `ListCursor` opacity preserved; the new `MoltZapNonAdvancingCursorError` is a `Data.TaggedError` (typed, not a throw). P2 boundary decode hardened (canonical-form check). P3 drain carries typed errors end-to-end. P4 n/a.

### Findings
- P1: 0. P2: 0. No findings.

My prior P3 (drain non-advancing-cursor guard) is resolved exactly as recommended — cursor-progress detection, not a page cap. No new findings.

Clean round.

APPROVE.

chughtapan:
## Verdict: APPROVE

**review-senior — 2nd consecutive clean on `e41e5f0c`** (no code change since clean #1; same SHA). This is NOT a rubber-stamp — I gave the three named subtle areas a fresh adversarial pass with throwaway empirical probes (all removed; tree clean). Reviewer: Claude (`/safer:review-senior`). Merge stays USER-GATED.

CI status: green (`ci`, `conformance`, `test` all SUCCESS; Mintlify skipped).

### Fresh adversarial pass on the three focus areas

**1. Keyset predicate ↔ ORDER BY correspondence under ms-truncation — PROVEN byte-identical.** Drove `keysetWhere` + `sortKeyExpr` + `paginate` against real PGlite with an adversarial `created_at` set: four rows truncating to the same `.500` ms but differing in µs (`.500123` / `.500999` / `.500000` / `.500500`), a `.501` next-ms row, a `.499` earlier-ms row, and a whole-second `01.000000` row. A limit-1 keyset walk produced a sequence byte-for-byte identical to the oracle full `ORDER BY date_trunc('milliseconds', created_at) DESC, id ASC` scan — no skip, no dup, exact order. The four same-ms rows are adjacent and ordered purely by `id ASC` (pure tie-break). Confirms: `date_trunc('ms', created_at)` (timestamptz) vs the ms `pos.sortKey` (text, coerced to timestamptz by Postgres) compares with identical ordering to the ORDER BY expression, in both the `<` branch and the `=`+`id>` tie-break branch. `(sortKey, id)` is a strict total order matching ORDER BY exactly.

**2. Cursor opacity boundary — airtight, no gap.** Probed the canonical re-encode check (`buf.toString("base64url") === cursor`) against six adversarial token variants: empty string, `=`-padded, standard-base64 alphabet (`+`/`/`), leading whitespace, trailing whitespace, and interior junk prefix/suffix (`!`/`@`). ALL rejected as typed `InvalidCursorError` (→ `InvalidParamsError` at the handler boundary); the legit canonical cursor still decodes. Every variant where Node's permissive base64url decoder would silently drop non-alphabet bytes and accept a non-canonical token is now closed. This is exactly the class of boundary bug a fresh pass should hunt; it is sealed.

**3. `drainPaginatedList` guard ↔ legit multi-page drain — no false-positive possible.** Re-derived the interaction: the guard trips only on `nextCursor !== undefined && seenCursors.has(nextCursor)`. A correct keyset server emits a strictly-advancing `(sortKey,id)` cursor per page (injective codec, unique-PK tie-break, keyset excludes emitted rows), so a cursor can repeat only if two pages end on the same row — impossible. An `undefined` nextCursor short-circuits the `&&` and ends the loop cleanly. The legit 3-page drain test passes with the guard active; the byzantine constant-cursor test (proven non-vacuous last round: unguarded → `RangeError` memory exhaustion; guarded → terminates in 2 calls, `catchAll` → `[]`) covers the other side. Chunking boundary re-checked: `AGENTS_LOOKUP_MAX_IDS = 100` equals `agents/lookup`'s wire `maxItems: 100`, and `slice(i, i+100)` stepping `i += 100` has no off-by-one at exact multiples (len 100 → one chunk; len 0 → no chunk; len 130 → 100 + 30).

### Re-confirm the whole (unchanged SHA)
Directory suite (3 tests) green; chunking constant agrees with the wire cap; no Record stragglers; opacity eslint guard fires (verified prior round, unchanged code). All prior-round verifications stand (server-core keyset+codec+contact 34 tests, openclaw 61 tests, divergence gate, builds, lint).

### Findings
- P1: 0. P2: 0. No findings.

Track A is at the merge bar (2 reviewers × 2 consecutive clean on the final SHA). Merge remains user-gated.

APPROVE.

chughtapan:
## ✅ Merge bar met — `e41e5f0c` (PARKED for user merge)

**N=2 × 2-consecutive-clean on the final SHA, CI green.**

| Gate | Status |
|---|---|
| CI (`ci` / `conformance` / `test`) | ✅ all green on `e41e5f0c` |
| codex (`gpt-5.5`/medium) | ✅ 2 consecutive clean (r5, r6) |
| review-senior | ✅ 2 consecutive clean (adversarial: keyset↔ORDER BY proven byte-identical vs PGlite; 6 cursor-opacity attack variants rejected; drain guard no-false-positive) |

**Resolved across the fix-rolls (each with a non-vacuous regression test):**
- P2 openclaw peer-directory truncation → consumer-side `drainPaginatedList` (+ chunked `agents/lookup`)
- P2 CLI `--json` dropping `nextCursor` → full result object
- P3 drain infinite-loop on byzantine cursor → cursor cycle-detection guard (typed fail, not a page cap)
- P2 NUL byte in `list-cursor.ts` (file was binary to git) → removed
- P2 permissive base64url accepting malformed cursors → canonical re-encode check (fail-closed → `InvalidParamsError`)

Branch rebased onto current `origin/main`; gates re-verified post-rebase. **Merge is user-gated — not merged, labels not advanced.**

chughtapan (commented):
## codex round 1 (SHA 03d3c6b05f3ee0027bb3204d52359a5de096537b)

**Verdict: changes_requested** — 1 P1 (verified reachable in production).

Independent review via OpenAI codex (`codex exec review --base main`, model gpt-5.5, reasoning=medium). One finding; grep-verified against HEAD before classifying.

### [P1] Unbound `sendRpc` method breaks OpenClaw directory listing in production

`packages/openclaw-channel/src/openclaw-entry.ts → listPeersEffect` passes `service.sendRpc` as a **bare callback** to both `drainPaginatedList(service.sendRpc, ContactsList, "contacts")` and `lookupAgentsInChunks(service.sendRpc, agentIds)`. This strips the `this` receiver.

Verified:
- The production service (`createGatewayService` → `new MoltZapService(...)` when `deps.createService` is absent) defines `sendRpc` as a **prototype method** (`packages/client/src/service.ts → MoltZapService.sendRpc`) whose body reads `this.client` inside `Effect.suspend`. Constructor does no binding; not an arrow class-field.
- Called as a bare reference, `this` is `undefined` (ESM strict mode). `drainPaginatedList` invokes `sendRpc(definition, params)` directly (`packages/client/src/pagination.ts`), so accessing `this.client` throws synchronously inside the suspend thunk → captured as an Effect **defect** (die), which `Effect.catchAll(() => Effect.succeed([]))` does NOT catch. `Effect.runPromise(listPeersEffect(...))` rejects → `directory.listPeers` rejects to the OpenClaw host. The directory is broken in normal operation.
- `lookupAgentsInChunks` has the identical defect at the second call site.

**Why CI is green (bug masked):** `openclaw-entry.directory.test.ts → startDirectoryGateway` injects `createService: () => ({ ...fixture.service, sendRpc: directorySendRpc })` where `directorySendRpc` is a standalone module function that never reads `this` (it closes over module-level state). The receiver-stripping is invisible to a `this`-free stub, so the pagination tests pass while production fails.

**Fix:** Wrap the method to preserve the receiver at both call sites, e.g.:
```ts
const send: typeof service.sendRpc = (definition, params) => service.sendRpc!(definition, params);
const contacts = yield* drainPaginatedList(send, ContactsList, "contacts");
...
const agents = yield* lookupAgentsInChunks(send, agentIds);
```
or `service.sendRpc.bind(service)`. Consider adding a directory test that uses a `this`-dependent `sendRpc` (or the real `MoltZapService` shape) so the unbound-method class can't regress silently.

---
CI: ci pass, conformance pass, test pass, Mintlify skipped (all green — consistent with the bug being masked by the `this`-free test stub).

chughtapan (commented):
## review-senior round 1 (SHA 03d3c6b05f3ee0027bb3204d52359a5de096537b)

Aggregate verdict: **APPROVE**. CI status: green (ci, conformance, test all pass). Build + lint + protocol/client/server-unit + agents-list integration + openclaw directory tests re-run locally — all green.

Composed gstack skills (`/review`, `/simplify`, `/codex`) are not installed in this environment; this is the safer-floor (P1-P4) + scope/merge-reconciliation review. Verdict tagged `DONE_WITH_CONCERNS` at the dispatcher boundary for that unavailability (per Unavailability rule); the underlying diff is APPROVE-clean with no P1/P2.

### Merge reconciliation (D3 cutover) — clean
The merge of origin/main (51 commits) did not drop pagination on any in-scope list RPC nor revert a D3-era change. All three Track-A list RPCs keep the convention: `AgentsList`/`ContactsList`/`TaskList` each carry `{ limit?: ListLimitSchema, cursor?: listCursorSchema() } → { <collection>, nextCursor?: listCursorSchema() }`. The 13 merge-conflict files were verified against `origin/main`: `tasks.handlers.ts` keeps the new `nextCursor` + `InvalidCursor→InvalidParamsError` mapping; `message.service.ts` keeps the D3 snowflake `sinceSeq`/`hasMore`; `messages.ts` keeps the `sinceSeq` model; `openclaw-entry.ts` keeps the `drainPaginatedList` relocation. No silent drop.

### Convention consistency (spec #694) — consistent
`task/list`, `contacts/list`, `agents/list` all use the shared `ListLimitSchema` (`DEFAULT_PAGE_LIMIT=50` / `MAX_PAGE_LIMIT=200`) and branded `listCursorSchema()` in + `nextCursor` out. Server codec (`db/list-cursor.ts`) is the single producer/decoder; the new eslint cursor-opacity guard bans `atob`/base64url `Buffer.from` elsewhere. `MessagesList` correctly stays on the legacy snowflake model (spec records it out-of-scope); `TaskConversationList` stays on the raw-string legacy cursor (Track B retires it). Type canary `identity-list-pagination.types-check.ts` locks the wire shapes incl. the Record→Array break.

### Client drainer relocation — correct (and a fix)
`drainPaginatedList` + `NonAdvancingCursorError` moved openclaw-channel → `@moltzap/client`, re-exported from the client index, re-imported by openclaw. `listPeersEffect` now drains ALL `contacts/list` pages (previously a single first-page call on main — Track A bounded the page to 50, so the prior code silently lost contacts past 50; the drain fixes that). Multi-page + byzantine cycle-guard both covered by `openclaw-entry.directory.test.ts` (130 contacts / 50-per-page = 3 calls; constant-cursor terminates at 2 calls).

### Craft (P1-P4) — green
- P1: branded `ListCursor`, `Equal<>` type canaries, `NonEmptyArray`-style invariants in types.
- P2: cursor decoded through a schema-style codec with a canonical-re-encode check (rejects Node's permissive base64url trailing-junk); decode failures are typed `InvalidCursorError`.
- P3: `InvalidCursorError` (tagged) → `InvalidParamsError` at each of the 3 handlers; no raw throws on the hot path; no silent catches.
- P4: `absurdTaskStatus(never)` exhaustiveness in `TaskService.loadOpenTask`.
- No raw SQL: Kysely throughout; the ms-truncation uses `eb.fn("date_trunc", ...)` (builder), not a string literal.

### Tests — non-vacuous
`list-cursor.test.ts` fast-check roundtrip + paginate-slice + nextCursor-points-at-limit-row properties; `task.service.test.ts` sub-millisecond keyset regression (seeds 3 rows sharing a ms differing in µs, walks limit=1 pages, asserts no skip/dup — fails against a non-truncated keyset); `pagination.test.ts` drain + cycle-guard. Conformance: `AgentsList` Record→Array divergence proof updated; `MODEL_METHOD_OUTCOMES[AgentsList]` ok→uncertain (server-validated cursor makes a schema-valid-but-undecodable token yield InvalidParams) — matches every other cursor-bearing list RPC.

### Findings (non-blocking)

**P3 — PR-body summary contradicts the diff on `MessagesList`.** The PR body's plan-anchor table says "`MessagesList` left as-is (recorded)", but the diff DOES change it: `limit` param now uses `ListLimitSchema` and the server clamp moves `MAX_MESSAGE_HISTORY_LIMIT=100` → `MAX_PAGE_LIMIT=200` (`message.service.ts → listMessages`, the `Math.min` clamp). Net effect: a `messages/list` caller requesting `limit:200` now gets up to 200 rows instead of being silently truncated to 100. This is BENIGN and a net improvement (the wire schema already advertised `maximum:200`, so this fixes a latent wire/server mismatch), and the CHANGELOG documents it correctly under "Cap raise — message history". Only the PR summary line is stale. No test pins the new 200 boundary for `messages/list` (`message-history.test.ts` uses `limit:100`); a boundary assertion would lock the fix.

**P3 — `listGroupsEffect` does not drain (latent truncation, pre-existing).** `openclaw-entry.ts → listGroupsEffect` still issues a single `sendRpc(TaskConversationList, {})` first-page call, while `listPeersEffect` was upgraded to drain. `TaskConversationList` is also page-bounded, so `listGroups` enumerates only the first page of conversations. Pre-existing (unchanged from main) and `TaskConversationList` paging is Track B's territory, so out-of-scope here — but the newly-relocated `drainPaginatedList` is now the natural fix and could be applied when Track B reshapes that RPC.

Aggregate rule: zero P1/P2 across the safer-floor checks + scope/merge audit; CI green → APPROVE (tagged DONE_WITH_CONCERNS only for the gstack-skill unavailability, not the diff).

chughtapan (commented):
## codex round 2 (SHA `2fb01eb6593e4970c9fc080987e4a6acc2c30164`)

**Independent reviewer:** OpenAI codex (`gpt-5.5`, `model_reasoning_effort=medium`) via `codex exec review --base main`.

### Verdict: CLEAN — no P1/P2 findings

codex's final verdict, verbatim:

> The pagination changes are consistently threaded through protocol schemas, handlers, services, and the updated consumers. Relevant typechecks and targeted tests pass, and I did not find a discrete introduced bug that would block correctness.

### What codex actually exercised (not just diff-skim)

- Read the full new surface: `list-cursor.ts` (codec + `keysetWhere` + `paginate`), `protocol/pagination.ts`, `client/pagination.ts` (`drainPaginatedList`), the `listCursorSchema`/`ListCursor` brand, and the `identity-list-pagination.types-check.ts` canary.
- Typechecked clean: `@moltzap/server-core`, `@moltzap/client`, `@moltzap/openclaw-channel` (all `tsc --noEmit` succeeded).
- Ran the targeted suites green: `src/db/list-cursor.test.ts` (7), `src/identity/services/contact.service.test.ts` (13), `src/task/services/task.service.test.ts` (14) — **34 passed**, including `TaskService > paginates sub-millisecond created_at ties without skips or dups`.
- Read the openclaw `listPeersEffect` cutover (now drains via `drainPaginatedList` + `lookupAgentsInChunks`) and the `model-equivalence.ts` confident-set note (`agents/list` correctly dropped from K because a schema-valid-but-undecodable cursor now yields `InvalidParamsError`).

### Independent verification (this reviewer, grep + read)

I re-read the load-bearing logic to confirm codex didn't rationalize past a hypothetical:

- `paginate` uses the canonical `limit+1` overflow pattern; `nextCursor` is emitted **iff** `rows.length > limit`. Correct.
- `decodeListCursor` fails **closed** on non-canonical base64url — the `cursor + "!"` Node-decoder-leniency case is explicitly defeated by the re-encode equality check (`buf.toString("base64url") === cursor`), then validates ISO-8601 canonicality + UUID shape. Maps to `InvalidParamsError` at the handler boundary. Sound.
- `keysetWhere` is a correct keyset predicate for `(sortKey DESC, id ASC)` with proper tie-break, and `sortKeyExpr` ms-truncates `created_at` in **both** the order-by and the predicate, so sub-ms ties are not skipped (the integration/service test above proves this end-to-end).
- `drainPaginatedList` cycle guard (`seenCursors`) checks-then-adds; no false positive on the first legitimate advance, terminates on a repeated cursor with typed `NonAdvancingCursorError`. Sound.

### P3 (advisory, non-blocking)

- **P3 — codex's first `vitest` invocation targeted a stale path** (`src/integration/identity/agents-list.test.ts`; the file actually lives at `src/__tests__/integration/identity/agents-list.test.ts`). That run was abandoned (codex's exec hit a closed-stdin error), and codex re-ran the 3 valid suites green and read the integration test source directly instead. Per repo convention, `src/__tests__/integration/` is excluded from the standard typecheck/run anyway, so the integration assertions were verified by source-read, not execution. No code defect — noting for completeness so the integration coverage is understood to be eyeballed, not re-run, in this round.

### CI

| Check | Status |
|---|---|
| `ci` | pass |
| `conformance` | pass |
| `test` | pass |
| Mintlify Deployment | skipping |

All green on `2fb01eb6`.

chughtapan (commented):
## review-senior round 2 (SHA 2fb01eb6593e4970c9fc080987e4a6acc2c30164)

Aggregate verdict: **APPROVE** (CI status: green — `ci` pass, `conformance` pass, `test` pass; Mintlify skipped).

Composed-skill note: running in fallback mode (no gstack composed skills in env), so the four safer craft checks (P1-P4) plus scope/test alignment are the aggregate. The prior round's `/review` codex pass already caught the real keyset microsecond bug (recorded in PR body) and it is fixed + regression-tested.

### Merge-resolution scrutiny (D3 reconciliation)
- No list RPC lost pagination in the merge. `TaskList`/`ContactsList`/`AgentsList` all carry `{ limit?, cursor? } → { <collection>, nextCursor? }`. `MessagesList` (`sinceSeq`/`hasMore`) and `TaskConversationList` (`{items, nextCursor?}`, `Type.String` cursor) are the recorded pre-existing schemes left as-is per spec #694 §5 (Track A scope) — not a silent drop.
- No D3-era change reverted: `MessagesList` now imports the shared `ListLimitSchema` (replacing the inline `maximum:200`), and `message.service.ts` swaps its local `DEFAULT/MAX_MESSAGE_HISTORY_LIMIT` for the protocol `DEFAULT_PAGE_LIMIT`/`MAX_PAGE_LIMIT` — a tightening, not a revert.

### Convention consistency (spec #693 Decision 1, #694)
- `TaskList`/`ContactsList`/`AgentsList`: `cursor` in (branded `listCursorSchema()`), `nextCursor` out (branded). `AgentsList` Record→`AgentCard[]` migrated across all 6 sites (descriptor, handler, CLI, integration test, conformance emitter, divergence proof). Bounded by `ListLimitSchema` (DEFAULT_PAGE_LIMIT=50, MAX_PAGE_LIMIT=200), single source in `pagination.ts`.
- Keyset `(created_at DESC, id ASC)` with ms-truncated `date_trunc` in BOTH order-by and `keysetWhere` — the microsecond-skip fix is correct and matches in column resolution and cursor sortKey.

### Client drainer relocation
- `drainPaginatedList` + `NonAdvancingCursorError` + `SendRpcFn` live in `@moltzap/client` (`pagination.ts`), wire-generic over `RpcDefinition`, exported from `client/src/index.ts`. Cycle guard is a `Set<string>`-backed non-advancing check, not a page cap. Correct home (channels/CLIs reuse it; openclaw is first consumer).

### Round-2 fix (`listPeersEffect` receiver bind)
- Real bug: `service.sendRpc` passed unbound to `drainPaginatedList`/`lookupAgentsInChunks` strips the receiver; the prototype method's `this.client` read inside `Effect.suspend` dies as a TypeError defect that `catchAll` (failure-channel) cannot absorb, so `directory.listPeers` rejects to the host. Fixed by `bind(service)` once. Regression test is non-vacuous: injects a `this.live`-reading `sendRpc` mirroring `this.client`; rejects without the bind, resolves with it.

### Craft (P1-P4)
- **P1 (types):** branded `ListCursor` makes cursor opacity structural; eslint `no-restricted-syntax` bans `atob`/base64url `Buffer.from` outside `db/list-cursor.ts` (server lint green). `agents: AgentCard[]` positive canary + every consumer compile proves the Record shape is gone.
- **P2 (boundaries):** cursor decoded via `decodeListCursor` (typed, with canonical-base64url re-encode check that rejects Node's permissive trailing-junk). `limit` capped at the AJV boundary (`validateParams` in `rpc-groups.ts` rejects `>200` before the service).
- **P3 (typed errors):** `InvalidCursorError` (TaggedError) → `InvalidParamsError` at every handler via `Effect.catchTag`. No raw throws on the new path, no silent catches. No raw SQL (Kysely builder, `date_trunc`/`eb.fn`; test seeds insert via `insertInto().values()`).
- **P4 (exhaustiveness):** no new union switches; existing `absurdTaskStatus` retained.

### Tests
- `list-cursor.test.ts`: fast-check roundtrip + typed-failure property + the trailing-junk regression. `task.service.test.ts`: sub-millisecond-tie pagination walk (exercises the keyset fix, fails pre-fix). `pagination.test.ts` (client): drain-to-tail + cycle-guard. Conformance `{agents:{}}`→`{agents:[]}` and the model-equivalence K/dispatch outcome flip for `agents/list` (now cursor-bearing → uncertain) are correct. New `identity-list-pagination.types-check.ts` canary locks the wire shapes. All suites green locally (client 307, protocol 197, server unit 34, openclaw 4) + CI test pass.

### Non-blocking (P3)
- `pagination.ts` JSDoc says "the server clamps to this value defensively," but `TaskService.list`/`ContactsService.list`/`agentsListPage` apply only `?? DEFAULT_PAGE_LIMIT` with no `Math.min(limit, MAX_PAGE_LIMIT)`; only `message.service.ts` clamps. Harmless — the AJV wire boundary already rejects `limit > 200` — but the doc claim is inaccurate for these three services. Either drop the "server clamps defensively" sentence or add the symmetric clamp. Not merge-blocking.

Zero P1/P2 findings. CI fully green. APPROVE.

chughtapan (commented):
## codex round 3 (SHA 2fb01eb6593e4970c9fc080987e4a6acc2c30164)

**Verdict: clean** — zero P1/P2 findings.

OpenAI codex (gpt-5.5, medium reasoning) reviewed the full diff against `main`. Final verdict:

> I did not find any discrete correctness issues introduced by this diff. The changed pagination paths typecheck, and the targeted unit tests for the new cursor codec plus task/contact service pagination pass in this environment.

### What codex exercised
- Read the cursor codec (`packages/server/src/db/list-cursor.ts`), protocol pagination schemas (`packages/protocol/src/pagination.ts`, `schema-primitives.ts`), the client drain helper (`packages/client/src/pagination.ts`), openclaw directory drain (`openclaw-entry.ts`), and all paginated handlers/services.
- Ran `tsc --noEmit` on `@moltzap/server-core` — **passed**.
- Ran the new unit suites — **34/34 passed**: `list-cursor.test.ts`, `task.service.test.ts` (incl. "paginates sub-millisecond created_at ties without skips or dups"), `contact.service.test.ts`.

### Independent verification (reviewer grep-check)
Spot-checked the three highest-risk areas; all behave as codex implies:
- **Cursor codec** — fails closed on tampered tokens (canonical-base64url re-encode check rejects `cursor + "!"`), validates ISO-8601 round-trip + UUID; ms-truncates `created_at` in BOTH order-by and keyset predicate so `(sortKey, id)` is a total order. Keyset `sortKey < k OR (sortKey = k AND id > id)` correctly matches `(created_at DESC, id ASC)`.
- **Drain loop** (`drainPaginatedList`) — terminates on `nextCursor === undefined`; `seenCursors` set + `NonAdvancingCursorError` guards against infinite loops on a misbehaving server.
- **paginate()** — standard `limit + 1` overflow technique; correct page slice + nextCursor derivation.

### Non-finding (environment artifact, NOT a code defect)
codex saw `agents-list.test.ts` (integration) fail with `EPERM` on `listen 0.0.0.0` / `connect 127.0.0.1` — the codex sandbox blocks socket binding. These same integration tests **pass in CI** (test job green). No action.

### CI
ci ✅ pass · conformance ✅ pass · test ✅ pass · Mintlify skipped.

chughtapan (commented):
## review-senior round 3 (SHA 2fb01eb6593e4970c9fc080987e4a6acc2c30164)

Fallback mode (composed gstack `/review` `/simplify` `/codex` not on disk in this dispatch env; the four safer craft checks P1-P4 carried by the aggregate). Read `git diff origin/main...HEAD` (51-commit main merge + the round-3 receiver-bind fix). Verdict: **APPROVE**. CI status: green (ci / conformance / test all pass).

### Scrutiny areas (all clear)

1. **Merge resolution preserved pagination + D3.** Handlers stay relocated under `identity/handlers/` (`agents-lookup.handlers.ts`, `contacts.handlers.ts`) and `task/handlers/` (`tasks.handlers.ts`); `TaskList` rename intact; `NonAdvancingCursorError` relocated to `@moltzap/client` `pagination.ts`. No list RPC silently dropped its `nextCursor`/`cursor`/`limit`. No D3-era change reverted.

2. **Convention consistent across the three target RPCs.** `AgentsList` / `ContactsList` / `TaskList` all use `{ limit?: ListLimitSchema, cursor?: listCursorSchema() } -> { <collection>: Array, nextCursor?: listCursorSchema() }`. `DEFAULT_PAGE_LIMIT=50` / `MAX_PAGE_LIMIT=200` single-sourced in `pagination.ts` and imported by every server clamp. Seq-cursor RPCs (`MessagesList`, `TaskConversationList`) correctly stay on their own cursor model and adopt only the shared `ListLimitSchema` — matches the recorded scope.

3. **Client drainer relocation correct.** `drainPaginatedList` is wire-generic, exported from `@moltzap/client`, cycle-guarded against a non-advancing server cursor. The round-3 fix in `listPeersEffect` (bind `service.sendRpc` receiver once before both drain consumers) is a real bug catch: a bare-reference prototype method strips `this`, dying with a `this`-undefined TypeError that the failure-channel `catchAll` cannot absorb. Regression test injects a receiver-dependent `sendRpc` (`makeReceiverDependentSendRpc`, reads `this.live`) and asserts `listPeers` resolves with all 130 contacts — non-vacuous.

4. **Conformance + integration + canaries cover pagination.** `identity-list-pagination.types-check.ts` locks all 4 wire invariants per RPC (agents Array-not-Record + cursor envelope; contacts cursor envelope; TaskList interim nextCursor). `MODEL_METHOD_OUTCOMES[AgentsList]` moved `ok -> uncertain` (a server-validated cursor makes a schema-valid-but-undecodable token yield `InvalidParamsError`) — self-consistent with every other cursor-bearing list RPC. Divergence-proof + client `rpc-semantics` emitters migrated `{agents:{}} -> {agents:[]}`. Codec property tests (roundtrip, typed-failure-or-valid, slice invariant, nextCursor-points-at-limit-row, trailing-junk canonical re-encode guard) + the sub-millisecond keyset regression both non-vacuous.

### P1-P4 craft notes
- **P1 Types.** Branded `ListCursor`; type-canary encodes the wire invariants as compile-time equalities. Clean.
- **P2 Boundaries.** Cursor decoded through `decodeListCursor` at the boundary with a canonical re-encode check (rejects Node's permissive base64url drop). Service tests run against real Postgres (testcontainers), not mocks, so the `date_trunc('milliseconds')` keyset coercion is exercised end-to-end.
- **P3 Errors.** Tagged `InvalidCursorError` -> `InvalidParamsError` at all three handlers. No raw throws in new code (the lone `throw` is the `absurdTaskStatus` exhaustiveness guard). No raw SQL in the new keyset path (`eb.fn`/`eb.or`/`eb.and` builder only); the eslint cursor-opacity guard bans `atob`/base64url `Buffer.from` outside `db/list-cursor.ts` and passes.
- **P4 Exhaustiveness.** `absurdTaskStatus(never)` on the task-status switch. Clean.

### Non-blocking note (informational)
`MessageService` history MAX clamp changed 100 -> `MAX_PAGE_LIMIT` (200), doubling the `messages/list` page ceiling. Defensible — the wire schema already permitted `maximum: 200` and the server was under-clamping — and `MessagesList` is recorded as in-scope-for-limit-only. Not a defect.

Local gates: build green; server-core 200, protocol 197, client 307, integration (agents-list 14 / contacts 6) all pass. Zero P1/P2 findings.

PR #680, opened 2026-05-22
refactor(server): canonical config — one file, one loader, one boot plan

chughtapan:
## Summary

Two commits collapse the entire config story into one file:

| # | Commit | LOC delta |
|---|---|---|
| 1 | Drop Ajv from the config loader; Effect Config does the same validation | −547 (schema.ts + tests + 2 deps) |
| 2 | Consolidate everything into src/config.ts | 5 files → 1, ~−220 net |

## Before

Five files spread across two folders, with four config shapes and four loader paths:

```
app/config.ts (216)             CoreConfig + LoadedConfig + ServerConfigLoader
config/schema.ts (283)           Ajv schema for the YAML shape
config/effect-config.ts (163)    Effect Config schema for the SAME YAML shape
config/loader.ts (352)           YAML pipeline (uses both schemas above)
config/runtime-config.ts (371)   Combiner: chains ServerConfigLoader + loadConfigFromFile
```

Plus 4 test files.

## After

```
src/config.ts (~700)     Everything: CoreConfig + StandaloneBootPlan + loadStandaloneConfig
src/config.test.ts (180) End-to-end loader tests
```

**One public type for callers** (`CoreConfig`), **one boot-plan type** (`StandaloneBootPlan`), **one loader function** (`loadStandaloneConfig`), **one error class** (`ConfigLoadError`).

`MoltZapAppConfig` lives internally as `YamlConfig` — operators write YAML, but no consumer references its TS type.

## What's gone

- `LoadedConfig` (env-only intermediate)
- `RuntimeProcessConfig` (combiner intermediate)
- `ServerConfigLoader` (env-only Effect)
- `MoltZapAppConfig` (public export → private)
- `runtimeConfigPath` Brand
- `RuntimeConfigSurfaceError`
- The whole `config/` folder
- Ajv schema validation (Effect Config validates the same shape)
- 2 npm deps (`ajv`, `ajv-formats`)

## Behavior preserved

- PGlite fallback when `DATABASE_URL` unset
- `CORS_ORIGINS` required outside dev mode
- Supabase URL rejected under `MOLTZAP_DEV_MODE=true`
- YAML `\${VAR}` interpolation
- `services.{sessions,contacts}: { type: webhook }` → adapter wiring

## Behavior dropped

- `regex:foo` CORS prefix parsing (the compiled regex patterns were never consumed by http-routes' allowlist — dead code)
- Ajv's polished error messages — Effect Config errors are simpler but adequate

## Trade-off

[...]

PR #679, opened 2026-05-22
refactor(server): rebalance misfiled handlers + adapter back-edges

chughtapan:
## Summary

Six mechanical folder-rebalance commits. Each moves a file (or
interface) to its natural domain home and updates the import sites.
No semantic changes; each commit tsc-clean + tests-pass.

| # | Move | Lines moved |
|---|---|---|
| 1 | \`task/handlers/presence.handlers.{ts,test.ts}\` → \`network/handlers/\` | 54 prod + test |
| 2 | \`task/handlers/contacts.handlers.ts\` → \`identity/handlers/\` | 103 |
| 3 | \`task/handlers/connect.handlers.ts\` → \`identity/handlers/\` | 221 |
| 4 | \`ContactService\` interface from \`app/app-host.ts\` → new \`identity/services/contact-policy.ts\` (fixes adapters→app back-edge) | 17 (new file) |
| 5 | \`WebhookSessionValidator\` from \`identity/services/session-validator.ts\` → new \`adapters/webhook-session-validator.ts\` | ~80 |
| 6 | \`runtime-surface/config.ts\` → \`config/runtime-config.ts\`; delete \`runtime-surface/\` folder | 371 |

## Why

Three RPC handler files lived in \`task/handlers/\` but their domain
services live elsewhere:

- **Presence** is a \`network/\` concern (\`presence.service.ts\` is in
  \`network/services/\`).
- **Contacts** is an \`identity/\` concern.
- **Connect** is the WS auth handshake — identity.

Two adapter impls were stranded:

- \`WebhookContactService\` (in adapters/) was importing
  \`ContactService\` interface from \`app/app-host.ts\` — adapters→app
  reverse-layer.
- \`WebhookSessionValidator\` was IN \`identity/services/\` despite being
  an adapter (uses \`WebhookClient\`); the impl belonged in adapters/.

And \`runtime-surface/\` was a one-file folder.

## Net effect

\`\`\`
                  before   after
task/             5671  →  5293  (−378, ~6.7% lighter)
identity/          859  →  1183  (+contacts+connect handlers + ContactService iface)
network/           846  →   900  (+presence handler)
adapters/          337  →   430  (+WebhookSessionValidator)
config/            798  →  1169  (+runtime-surface/config.ts)
runtime-surface/   371  →     0  (folder gone)
app/              3355  →  3331  (−ContactService iface)
\`\`\`

Fixes the adapters→app back-edge that architecture-lint would flag.

## Test plan

[...]

PR #676, opened 2026-05-21
chore(server): orphan-shim cleanup + dead public surface

chughtapan:
## Summary

Seven mechanical / structural cleanup commits stacked on #675. No
architectural changes — that's coming in a follow-up. Each commit is
independently reviewable and tsc-clean.

| # | Commit | Net lines |
|---|---|---|
| 1 | drop unused public subpaths + dead layer barrels (5 files deleted) | −183 |
| 2 | inline server-constants.ts (10 lines) | −10 |
| 3 | fold hooks.ts (35 lines) into types.ts | −15 |
| 4 | inline logging.ts wrappers Effect-native | −22 |
| 5 | merge dev.ts → standalone.ts; pnpm dev → tsx watch standalone.ts (PGlite fallback when DATABASE_URL unset) | −38 |
| 6 | inline runtime/direct-run helpers, delete runtime/ folder entirely | −10 |
| 7 | extract CoreConfig from types.ts to config.ts (boot input lives with config loader) | mixed |

## What this is NOT

This PR does NOT include the bigger architectural rewrites discussed:

- HTTP startup as Effect Layer (eliminating `Scope.extend(appScope)` + `runPromise(startup).catch`)
- Folder split `app/` → `app/` (host primitives) + `boot/` (composition)

Those are deferred to a follow-up PR stacked on this one — they touch
lifecycle semantics + many imports and deserve their own review.

## Test plan

- [x] `pnpm exec tsc --noEmit` clean on packages/server after every commit
- [x] `pnpm exec vitest run src/app src/standalone.test.ts` passes (30+ tests)
- [x] `pnpm exec agent-code-guard-knip --treat-config-hints-as-errors` clean
- [x] `pnpm lint` passes (pre-commit hooks gate every commit)
- [ ] `pnpm dev` boots with embedded PGlite when DATABASE_URL is unset
- [ ] `pnpm dev` connects to Postgres when DATABASE_URL is set
- [ ] `bin/moltzap-server` boots in standalone (unchanged code path)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #701, opened 2026-05-24
[impl-staff] consolidate server config into single src/config.ts

chughtapan:
Re-derives #680 (server-unified-config) against post-#676/#678 `main`. Targets `main`. Does NOT depend on #679 (held).

Plan/contract: `/home/tapanc/.claude/plans/680-redux.md`. The original `origin/refactor/server-unified-config` (2 commits: `9c57825f` consolidate, `4b66bdd3` drop Ajv) was authored on top of #679 and before #676's `CoreConfig` extraction, so a cherry-pick conflicts. Treated as INTENT; re-derived against the current layout. #679's `apps[]`/`registerConfiguredApps`/manifest-registration features are NOT pulled in (held with #679); `apps` is parsed and carried on the boot plan exactly as current `main` carries it, but not acted on.

## What changed
Folds the 5 current config sources (`app/config.ts`, `config/effect-config.ts`, `config/loader.ts`, `config/schema.ts`, `runtime-surface/config.ts`) into a single `packages/server/src/config.ts`. The `config/` and `runtime-surface/` folders are deleted. Public config surface narrows to `CoreConfig`, `StandaloneBootPlan`, `ConfigLoadError`, and one loader `loadStandaloneConfig`. Ajv is dropped (Effect `Config` validates the same shape); the dead `regex:` CORS prefix parsing is removed. Behavior is preserved otherwise.

## Traceability
| Artifact | Kind | Plan anchor |
|---|---|---|
| `packages/server/src/config.ts` | consolidated module | Plan "Scope IN" 1; AC 2, 3 |
| `CoreConfig` (moved from `app/config.ts`, #676) | public type | Plan "Scope IN" 2; AC 3 |
| `StandaloneBootPlan` | public type | Plan "Scope IN" 2; AC 3 |
| `ConfigLoadError` | public tagged error | Plan "Scope IN" 2; AC 3 |
| `loadStandaloneConfig` | public loader fn | Plan "Scope IN" 2; AC 3 |
| Ajv + ajv-formats removal | dep removal | Plan "Scope IN" 3; AC 4 |
| dead `regex:` CORS removal | simplification | Plan "Scope IN" 4; AC 5 |
| delete `config/`, `runtime-surface/` | folder removal | Plan "Scope IN" 5; AC 2 |
| repoint `standalone.ts`, `app/server.ts`, `app/http-routes.ts` | consumers | Plan "Scope IN" 6; AC 6 |
| `ARCHITECTURE.md` config/boot section | doc | Plan Guardrails; AC 8 |

## Scope
- New canonical module: `packages/server/src/config.ts` (consolidation of 5 prior files into 1).
- Public exports: `CoreConfig`, `StandaloneBootPlan`, `ConfigLoadError`, `ConfigLoadErrorKind`, `LoadStandaloneConfigInput`, `loadStandaloneConfig`.
- Deps removed: `ajv`, `ajv-formats` (server package.json + lockfile). Nothing else in `packages/server/src` imports them (`@sinclair/typebox` stays, used by transport/app code). `packages/protocol` keeps its own `ajv` dep — untouched.
- Tier (safer-diff-scope --head HEAD): reports `senior` (17 files, 2 modules, 6 exported-signature changes, `new_deps: 0`). The tool does not count a dependency *removal* as `new_deps`, so it lands at senior; the orchestrator dispatched staff because this is a public-surface change that removes a dependency. Not a stop per the implement-staff scope rules — noted here.

## CORS-regex decision
Removed (confirmed dead). The old `CorsConfig.patterns` compiled-regex array was never read outside `app/config.ts` itself; `corsOrigins` flows to `CoreConfig.corsOrigins: string[]` and `http-routes.ts → makeAllowedOriginsPredicate` does exact-match only (`includes("*")` / `includes(origin)`). `corsOrigins` is now exact-only. Not documented anywhere (`docs/snippets/env-vars-table.mdx` already describes `CORS_ORIGINS` as comma-separated, no `regex:` prefix), so no doc change needed.

## Public-surface blast radius
No in-repo consumer breaks. Cross-package imports of `@moltzap/server-core` are all via `@moltzap/server-core/test-utils` (re-exports `startCoreTestServer`, not config types) or doc comments. `arena` consumes only `@moltzap/protocol` + `@moltzap/client`, not server config. Confirmed by grep across `packages/*` and a clean full-workspace `tsc --noEmit`.

## Behavior preservation
Same fields parsed, same env precedence (env over YAML), same boot plan: PGlite vs Postgres selection (empty `databaseUrl` → PGlite), `CORS_ORIGINS` required outside dev, Supabase rejected under dev, YAML `${VAR}` env interpolation (non-mutating — regression covered by a test), missing-file tolerance for non-explicit paths (explicit path → fail loud). `ENCRYPTION_MASTER_SECRET` reads via `Config.redacted` + immediate unwrap so the Effect ConfigError formatter can't leak it.

## Tests
- `packages/server/src/config.test.ts` (new, 12 tests): env-only dev defaults, encryption surfacing, DATABASE_URL/PORT override, CORS parse, CORS-required-in-prod, Supabase rejection, invalid-port validation error, YAML env interpolation, sessions webhook binding, apps[] passthrough, and the non-mutating-reused-env regression.
- Deleted test files (`app/config.test.ts`, `config/loader.test.ts`, `config/schema.test.ts`, `runtime-surface/config.test.ts`) covered behaviors now exercised end-to-end through `loadStandaloneConfig`. The Ajv-specific niceties (unknown-field rejection, `format: uri`, keyword-by-keyword error formatting) are the accepted Ajv-drop tradeoff.
- Full server suite: 165/165 pass. Full-workspace `tsc --noEmit`, eslint, oxlint, knip all clean. Pre-commit hooks pass (no `--no-verify`).

## Simplify skips
- `/simplify` skill is not installed in this environment — skipped. A manual simplification pass found no dead code (knip clean); the file structure mirrors the user-authored original `9c57825f` decomposition.

## Review skips
- `/review`: no findings (0 critical, 0 informational). Scope CLEAN.

## Confidence
HIGH — full build/test/lint/typecheck/knip green; behavior traced field-by-field against the prior two-stage pipeline; public-surface blast radius grep-verified across all packages.


chughtapan:
## /safer:stamina reviewer — role: acceptance-vs-diff

**Verdict: APPROVE (pending CI)** — acceptance criteria met; no P2+ findings. Two P3 notes below.

Reviewed `gh pr diff 701` at `8ce2a70` against the plan ACs (`680-redux.md`). Verification done against the PR-head tree (`origin/impl/680-redux`), not the local worktree.

### Acceptance criteria

| AC | Status | Evidence |
|---|---|---|
| 2 — single `src/config.ts`; `config/` + `runtime-surface/` gone | PASS | `git ls-tree` on PR head: `packages/server/src/config/`, `runtime-surface/`, `app/config.ts` all absent; new `config.ts` present. |
| 3 — public exports `CoreConfig`, `StandaloneBootPlan`, `ConfigLoadError`, `loadStandaloneConfig`; `CoreConfig` lives here | PASS | All four exported from `config.ts`. `CoreConfig` interface is field-for-field identical to #676's `app/config.ts` (db, dbCleanup, encryptionMasterSecret, port, corsOrigins, registrationSecret, devMode, devModeUserId, sessionValidator, webhookClient, skipDefaultRegisterRoute, deliveryWebhook, spanProcessor). All `CoreConfig` importers (`app/http-routes.ts`, `app/server.ts`) repointed to `../config.js`. |
| 4 — Ajv removed; Effect Config validates | PASS | No `ajv`/`ajv-formats` import anywhere in `packages/server/src/` on PR head. Dropped from `packages/server/package.json` + `pnpm-lock.yaml`. (`protocol` package keeps ajv — genuinely uses it, out of scope.) `YamlConfigSchema` (Effect `Config`) validates types, port range `[1,65535]`, non-empty strings; `rejectsInvalidPort` test confirms a `port: -1` YAML fails with a `validation` `ConfigLoadError`. |
| 5 — dead `regex:` CORS parsing gone; `corsOrigins` exact-only | PASS | `makeAllowedOriginsPredicate` (`app/http-routes.ts`) is exact-only: `corsOrigins.includes("*") \|\| corsOrigins.includes(origin)`. On base main, the compiled `patterns: RegExp[]` array was projected into `CorsConfig.patterns` but `standalone.ts` only forwarded `corsOrigins.exact` to `CoreConfig.corsOrigins` — the patterns array never reached any consumer. Genuinely dead; removal justified. |
| 6 — all consumers repointed; no dangling config imports | PASS | No references to `config/effect-config`, `config/loader`, `config/schema`, `runtime-surface/config`, `app/config.js`, `loadConfigFromFile`, `loadRuntimeProcessConfig`, `ServerConfigLoader`, `runtimeConfigPath`, `RuntimeProcessConfig`, or `MoltZapAppConfig` anywhere in `packages/` on PR head. (`MoltZapConfig` matches in `client/src/cli/config.ts` are an unrelated CLI type — `{ serverUrl }` — not the server config symbol.) `ci` check is green, which means tsc is clean. |
| 7 — boot behavior unchanged | PASS | PGlite/Postgres selection preserved (`bootPlan.databaseUrl.length === 0` → PGlite with `pgliteDataDir`; else Postgres). Env precedence preserved: every `resolve*` reads `processEnv[KEY] ?? yaml.<field>`, env wins — matches old `runtime-surface` ordering (process-env applied after app-config). `encryptionMasterSecret`, `port`, `corsOrigins`, `registrationSecret`, `devMode`, `devModeUserId`, session/contact webhook bindings all carried through `StandaloneBootPlan` and fed to `makeCoreConfig`. Supabase-in-dev-mode rejection preserved (`resolveDatabaseUrl`). |
| 8 — arch/config docs updated | PASS | `ARCHITECTURE.md` §1 project-structure tree and §2 public-surface Config row both updated to the new file + exports. No `docs/architecture/NN-config.md` detail doc exists (config lives only in the index). No stale loader-symbol references in `docs/`, README, or package docs. README/docs reference only unchanged public surface (`CoreConfig`, `createCoreApp`, YAML keys) — no internal-path drift, so no path updates were required. |

(AC 1 build/lint/test green is CI's job — `ci` is SUCCESS; `test` + `conformance` still IN_PROGRESS at review time. See CI note.)

### Latent-bug fix — VERIFIED CORRECT

The implementer reported the original #680 `config.ts` dropped `callback_token` when `timeout_ms` was also present. Confirmed against `origin/refactor/server-unified-config:config.ts`:

```ts
const out = { type, webhook_url };
if (timeout_ms !== undefined) return { ...out, timeout_ms };          // early return drops callback_token
if (callback_token !== undefined) return { ...out, callback_token };
return out;
```

Both-present input took the first `return`, silently dropping `callback_token`. The PR fix (`config.ts:207`) handles both-present first:

```ts
if (timeout_ms !== undefined && callback_token !== undefined) {
  return { ...out, timeout_ms, callback_token };
}
if (timeout_ms !== undefined) return { ...out, timeout_ms };
if (callback_token !== undefined) return { ...out, callback_token };
return out;
```

All four field-presence combinations are now correct. No other webhook field is mishandled. Note (informational, not a defect): `callback_token` is **dead downstream** — `webhookBinding` (`config.ts:573`) projects only `url` + `timeoutMs` into `WebhookServiceBinding`, and `standalone.ts` reads only those. This matches base-main behavior (the field was parsed-but-unconsumed there too), so the fix is behavior-preserving on the observable boot path while correctly preserving the parsed shape. The fix itself is correct.

### Craft (P1–P4)

- P1: discriminated `YamlServiceConfig` union, `WebhookServiceBinding` interface. Clean.
- P2: YAML document decoded through `YamlDocumentSchema` (Schema.Record) at the file boundary; env read through `Config`. Clean.
- P3: `ConfigLoadError` is `Data.TaggedError`; loader path has no raw `throw` (`parseYaml` wrapped in `Effect.try` with typed catch). Clean.
- P4: `formatConfigError` uses `Match.discriminatorsExhaustive("_op")` over the ConfigError tree. Clean.

### P3 findings (resolve-but-non-blocking)

1. **`moltzap.example.yaml:46` still ships `log_level: info`.** `log_level` was modeled (and accepted) by the old Ajv schema but never consumed anywhere on main. The new `YamlConfigSchema` drops it entirely. Because Effect `Config.all` ignores undeclared keys (it has no `additionalProperties: false` equivalent), the line is now silently tolerated rather than read — no boot regression, but the example advertises a key the schema no longer models. Drop the line.

2. **Ajv-drop tradeoff is slightly broader than "less polished error messages."** The old Ajv schema used `additionalProperties: false` everywhere and actively rejected unknown/typo fields (e.g. the retired `seed:` block, a misspelled `databse:`). Effect `Config.all` silently ignores unknown keys, so a typo'd config key now no-ops instead of failing loudly. This is within the spirit of the user-accepted Ajv-drop (AC 4 only requires validating the declared shape, which the new code does), but the strictness loss is wider than the error-message polish the plan called out. Acceptable as-is; flagging so it's a conscious record, not a surprise.

### CI

`ci` SUCCESS (tsc/lint green → AC 1 build+lint satisfied, AC 6 tsc-clean confirmed). `test` and `conformance` IN_PROGRESS at review time. No APPROVE should finalize-to-merge until both clear. Verdict caps at APPROVE-PENDING-CI on the runtime checks.

---
STATUS: DONE_WITH_CONCERNS (acceptance role clean / no P2+; concerns: 2x P3 + CI test+conformance pending)
Confidence: HIGH (verified against PR-head tree, original #680 branch, and base main; webhook fix traced end-to-end)

chughtapan:
## /codex adversarial review — config consolidation (PR #701)

**STATUS: DONE_WITH_CONCERNS** — verdict: **2 blocking concerns (P2)**, plus a cluster of P1 validation-strictness regressions that are arguably in-scope-of-"drop Ajv" but undefended by tests.

Cross-model (codex `gpt-5`-family, medium reasoning, self-contained prompt). Every codex finding was cross-checked against the actual `gh pr diff 701` and main's behavior. Phantoms dismissed below.

### Method
Diffed new `src/config.ts` against the 5 deleted sources (`config/schema.ts` [Ajv], `config/effect-config.ts`, `config/loader.ts`, `runtime-surface/config.ts`, `app/config.ts`) and the `standalone.ts` consumer. Confirmed which old constraints were *tested behaviors* via deleted `config/schema.test.ts`.

---

### Blocking (P2)

**[P2-A] A typo'd encryption key now silently disables at-rest encryption.**
`encryption: { masterSecret: "..." }` (camelCase typo for `master_secret`).
- Main: Ajv had `additionalProperties: false` on the encryption object AND `master_secret` was **required** (`Type.String`, not `Type.Optional`). Both the unknown key and the missing required field were rejected. Tested in deleted `config/schema.test.ts` ("rejects unknown nested fields", "Missing required field").
- PR: `YamlConfigSchema → encryption: opt(Config.all({ master_secret: opt(nonEmptyString) }))`. Effect `Config` ignores unknown keys, and `master_secret` is optional. So the block is accepted, `master_secret` resolves to `undefined`, and `assembleBootPlan` returns `encryptionMasterSecret: undefined` → **encryption layer silently off, messages stored plaintext**.
- This is a *behavior* change (silent security downgrade), not the "less-polished error messages" tradeoff the plan accepted (plan L30-33, L46). Crosses the "behavior-preserving" guardrail.

**[P2-B] `encryption: {}` (empty block) now accepted, was rejected.**
Same root cause: `master_secret` went from required-when-present to optional. Operator who writes an empty/partial encryption block gets no error and no encryption. `config.ts → YamlConfigSchema, encryption section`.

---

### P1 — validation-strictness regressions (tested behavior on main, no replacement test added)

These were deliberate, tested Ajv constraints (deleted `config/schema.test.ts`). The plan accepted "drop Ajv → less polished error messages" but did NOT scope away *rejection* of malformed input. New suite (`config.test.ts`, 8 cases) adds zero negative tests for any of these. Recommend either restoring the checks as `Config.validate`/`Schema` refinements, or getting explicit user sign-off that silent-accept is acceptable and documenting it.

- **[P1-1] `additionalProperties: false` gone everywhere.** Unknown keys silently pass. e.g. `server: { port: 3000, prt: 4000 }` (typo) is accepted, real `port` ignored if mis-typed. Main rejected (tested: "rejects unknown top-level fields" / "rejects unknown nested fields" / "rejects retired `seed` block"). `config.ts → YamlConfigSchema`.
- **[P1-2] `webhook_url` URI-format validation gone.** `services.sessions.webhook_url: "not a url"` accepted; only `nonEmptyString` now. Main: Ajv `format: "uri"` (tested "validates webhook_url format"). `config.ts → YamlWebhookServiceConfig`.
- **[P1-3] `timeout_ms` `minimum: 100` gone.** `timeout_ms: 0` accepted and passed straight to `WebhookSessionValidator` (the `?? DEFAULT_WEBHOOK_TIMEOUT_MS` only fires on `undefined`, not `0`). Main rejected sub-100. `config.ts → YamlWebhookServiceConfig`.

---

### P3 / informational

- **[P3] `log_level` enum validation gone.** `log_level: "verbose"` now silently accepted (key dropped from schema entirely). `log_level` had zero runtime consumers on main and zero now, so no behavior change beyond message — informational.
- **[P3] `callback_token` parsed but unused — NOT a regression.** Codex flagged this as field loss; dismissed. `callback_token` was dead config on main too (`git grep` finds no non-config/test consumer on either branch; `webhookBinding`/old standalone only ever carried url+timeout). The new 4-branch `Config.map` correctly preserves it at parse time (fixes the latent old-mapper drop bug) — but `webhookBinding` projects only `{ url, timeoutMs }`, same as main. Behavior-preserving. The implementer's "fix" is correct but cosmetic since nothing downstream reads the field.
- **[Secrets] Clean.** `ENCRYPTION_MASTER_SECRET` is read via `Config.redacted` then unwrapped into the snapshot — same operational exposure as main's `ServerConfigLoader` (which also unwrapped via `Redacted.value`). No new logging/leak path in error messages (`makeInvalidEnvError`/`formatConfigError` emit key paths + messages, not secret values). No regression.

---

### Phantoms dismissed (codex over-flagged against env path)

- **PORT range via env (`PORT=70000`):** codex called this P1. DISMISSED — env `PORT` was **never** range-checked on main either (old `runtime-surface parseIntegerEnv` had no range check; `ServerConfigLoader` used bare `Config.integer("PORT")`). Pre-existing, not introduced here. YAML `server.port` range **is** still enforced (new `YamlConfigSchema` keeps `portNumber [1,65535]`, and `config.test.ts` "rejects an out-of-range port" covers it).

---

### Boot-plan parity (focus #4): holds.
PGlite-vs-Postgres selection (`databaseUrl.length === 0`), `pgliteDataDir`, env-over-YAML precedence for `DATABASE_URL`/`PORT`/`CORS_ORIGINS`, dev-mode `["*"]` CORS default, and the Supabase-under-dev-mode guard all map 1:1 from the old `runtime-surface` + `ServerConfigLoader` paths. The `services.sessions` key is now read correctly (old `effect-config.ts` parsed `services.users` while `standalone.ts` read `services.sessions` — a latent mismatch the consolidation incidentally fixes; verify intended).

### Recommendation
Recommendation: Restore the encryption `master_secret`-required + unknown-key rejection (P2-A/P2-B) before merge because a camelCase typo silently turning off at-rest encryption is a behavior change beyond the "drop Ajv = softer error messages" tradeoff the plan signed off on, whereas the URI/timeout/additionalProperties P1s are defensible-as-scoped if the user explicitly accepts silent-accept and the team adds negative tests to lock the new contract.

<sub>codex medium reasoning, ~30.7k tokens; prompt self-contained (no repo grep). Findings cross-checked against `gh pr diff 701` @ 8ce2a70.</sub>

chughtapan:
## `/review` — structural-diff (stamina reviewer) — PR #701 unified config

**Verdict: DONE_WITH_CONCERNS** — consolidation is structurally sound. No silently-lost field that breaks the boot path. All carried fields, the boot plan, env precedence, PGlite/Postgres selection, and the `callback_token`/`timeout_ms` fix verified. Three P3 (non-blocking) concerns around validation relaxation + a test gap. **No P0/P1/P2.**

Verified locally at PR HEAD `8ce2a700`: full-workspace `pnpm -r build` clean (server-core tsc green), `@moltzap/server-core` test suite **165/165 pass** including `src/config.test.ts` (12 tests). Folders `config/` and `runtime-surface/` fully gone; `app/config.ts` deleted; zero dangling imports of deleted modules; no cross-package consumer of server config types; `ajv`/`ajv-formats` removed from package.json + lockfile and no longer imported in server src (`@sinclair/typebox` correctly kept).

### Field-by-field carry-over (5 deleted sources → `src/config.ts`)
Every field/default/env-var carried: `server.port` (1–65535 range validate preserved), `cors_origins`, `database.url`/`data_dir`, `services.sessions`/`contacts` (webhook|in_process), `registration.secret`, `dev_mode.enabled`/`user_id`, `apps[]`. Env precedence (env > YAML), `${VAR}` interpolation (non-mutating, regression-tested), missing-file tolerance (explicit path → fail loud), Supabase-under-dev rejection, `ENCRYPTION_MASTER_SECRET` via `Config.redacted` — all preserved. The consolidation also **fixes a latent bug** in the old `effect-config.ts`: its `ServicesSection` keyed `Config.nested("users")` instead of `sessions` — the new schema uses `sessions` correctly.

### `callback_token`/`timeout_ms` bug fix — CONFIRMED REAL
The original #680 (`9c57825f`) had:
```js
if (timeout_ms !== undefined) return { ...out, timeout_ms };   // early-return drops callback_token
if (callback_token !== undefined) return { ...out, callback_token };
```
Both-present → first branch returns and `callback_token` is dropped. PR #701 fixes it with an explicit both-present branch first (`config.ts → YamlWebhookServiceConfig` map). Correct, and the single-field/neither cases are preserved. Note: `callback_token` is **dead** in both old and new — `WebhookServiceBinding` only carries `url`+`timeoutMs`, and grep shows no consumer anywhere. So this fixes parse-fidelity for an unused field. Defensible, but see P3-c.

### Boot plan (`standalone.ts`) — correct
`loadStandaloneConfig` produces the `StandaloneBootPlan` `standalone.ts` consumes. PGlite vs Postgres selection preserved exactly (`databaseUrl.length === 0` → PGlite with `pgliteDataDir`, else Postgres). `sessionWebhook`/`contactWebhook` bindings, `devModeUserId` (random UUID fallback), `registrationSecret`, `corsOrigins` (now exact-only `string[]`), `encryptionMasterSecret` all wired through. Dead CORS-regex (`regex:` prefix, `CorsConfig.patterns`) removal confirmed safe — `http-routes.ts → makeAllowedOriginsPredicate` is exact-match only.

---

### P3 (resolve-but-non-blocking)

**P3-a — Ajv→Effect Config validation gap (acknowledged tradeoff).** (confidence 9/10) `config.ts`. The old pipeline ran Ajv (`schema.ts`) *before* Effect Config. Ajv enforced things Effect Config does not: `additionalProperties:false` (unknown-field rejection at every level), `webhook_url` `format:"uri"`, `timeout_ms` `minimum:100`, the `log_level` enum, and the retired-`seed`-block rejection. These had explicit tests (`schema.test.ts`: "rejects unknown top-level fields", "validates webhook_url format", "validates log_level enum", "rejects retired seed block") that are now deleted. The PR body explicitly accepts this ("the accepted Ajv-drop tradeoff"). Practical impact is low (typos in YAML now silently ignored instead of erroring) but the "Behavior preserved otherwise" claim overstates parity — the validation surface genuinely narrowed. Recommend: if loud-config-error UX matters, file a follow-up to re-add unknown-field rejection via an Effect `Schema.Struct` boundary decode (cheap, no Ajv).

**P3-b — `encryption.master_secret` relaxed from required to optional.** (confidence 8/10) `config.ts → YamlConfigSchema`, `master_secret: opt(nonEmptyString("master_secret"))`. Old `effect-config.ts` (`EncryptionSection`) and Ajv both **required** `master_secret` when an `encryption` block was present. Now `encryption: {}` is silently accepted → `encryptionMasterSecret` is `undefined` → `server.ts` disables envelope encryption (plaintext at rest) instead of failing loud. Fail-open on a security-relevant field. Edge case (who writes an empty encryption block), but worth a one-line `Config.validate` or doc note. Recommend: keep `master_secret` required-when-`encryption`-present, or document the relaxation.

**P3-c — zero test coverage for the `callback_token` both-present fix.** (confidence 9/10) `config.test.ts`. The PR calls out the both-present `callback_token`+`timeout_ms` drop as the headline bug fix, but no test exercises it (no `callback_token` anywhere in the new test file), and `contactWebhook` binding has no test either (only `sessionWebhook`). A 5-line test feeding `type: webhook` + both fields and asserting the parsed shape would lock the fix against regression. Recommend: add it.

---

`log_level: warn` is written into generated YAML by `claude-code-channel` and `nanoclaw-channel` integration global setups — now an unrecognized-but-tolerated field (Effect Config ignores unknown keys). Functionally fine (server never consumed it; client reads `MOLTZAP_LOG_LEVEL` env), no break — flagged only as the most visible instance of P3-a's relaxed unknown-field handling.

Scope: CLEAN. Intent (consolidate 5 sources → 1, drop Ajv, drop dead CORS-regex, delete folders) matches the diff exactly; no scope creep.

— structural-diff reviewer (Claude), stamina dispatch

chughtapan:
## Stamina verdict: ESCALATED — validation-semantics regression (encryption fail-open). Consolidation otherwise clean.

**Mode:** pr | **N:** 3 (source: table — tier=senior, exports=6 → public surface) | **CI:** green (but CI does NOT catch this — the negative tests that would were deleted with `config/schema.test.ts`).

### Dispatch set

| Role | Skill | Verdict | Comment |
|---|---|---|---|
| acceptance-vs-diff | `/safer:review-senior` | DONE_WITH_CONCERNS | https://github.com/chughtapan/moltzap/pull/701#issuecomment-4529532022 |
| structural-diff | `/review` | DONE_WITH_CONCERNS | https://github.com/chughtapan/moltzap/pull/701#issuecomment-4529542813 |
| adversarial / cross-model | `/codex` | DONE_WITH_CONCERNS (blocking) | https://github.com/chughtapan/moltzap/pull/701#issuecomment-4529534306 |

### Unanimously verified clean
Consolidation complete + behavior-preserving for everything except validation: all 5 sources folded into `src/config.ts`, both folders gone, `CoreConfig` field-identical, every field/default/env-var/`${VAR}` interpolation/PGlite-vs-Postgres selection/env precedence preserved, Ajv removed, dead CORS-regex removal justified, `callback_token` bug-fix correct (though that field is dead downstream). Consolidation incidentally fixes a latent `users`→`sessions` key bug. No secret leak. CI green.

### Blocking finding (must fix + re-review)

**Validation-semantics regression from the Ajv→Effect Config swap.** Negative behaviors enforced by Ajv on main (tested by the now-deleted `config/schema.test.ts`), not re-asserted:
- **Encryption fail-open (the severe one, codex P2 / review P3-b):** `encryption.master_secret` relaxed from required-when-present to optional, and unknown keys no longer rejected. So `encryption: { masterSecret: "..." }` (camelCase typo) or `encryption: {}` → `encryptionMasterSecret: undefined` → **at-rest encryption silently disabled, plaintext storage.** Beyond the accepted "softer error messages" tradeoff — this is weaker validation semantics with a data-protection consequence.
- **Lower-severity cluster (codex/review):** `additionalProperties: false` (unknown-key rejection), `webhook_url` URI format, `timeout_ms` min:100, `log_level` enum, retired-`seed` rejection — all dropped.
- No test coverage for the `callback_token` both-present fix.

### Next route
Encryption fail-open → MUST fix (restore required `master_secret` + unknown-key rejection for the encryption block). The lower-severity validation cluster → user decides full-parity-restore vs accept-softer-with-negative-tests. Then re-review + `/document-release` → merge.

chughtapan:
## Targeted re-review: validation regression fix (role: acceptance-vs-diff)

**Scope:** Confirm the blocking validation regression (encryption fail-open + dropped Ajv structural checks) is fixed with adequate, non-vacuous test coverage. Diff reviewed: `8ce2a700..97173aae`.

### Verdict: SHIP — regression fixed, parity complete, tests proven non-vacuous

---

### 1. Parity complete (P0-clear)

`MoltZapConfigShape` in `config.ts` is a faithful reconstruction of `main`'s `config/schema.ts → MoltZapConfigSchema` (the same TypeBox object that was fed to Ajv). Every rejection enforced by the deleted `config/schema.test.ts` is re-enforced:

| Ajv enforced | TypeBox mechanism |
|---|---|
| `master_secret` required when `encryption` present | `encryption` → required `master_secret` (minLength 1) |
| camelCase `masterSecret` typo (the security fail-open) | `additionalProperties: false` + required key |
| unknown top-level / nested keys | `additionalProperties: false` at every object level |
| retired `seed` block | rejected by top-level `additionalProperties: false` |
| `webhook_url` URI format | `format: "uri"` + `FormatRegistry.Set("uri", URL.canParse)` |
| `timeout_ms` min 100 | `Type.Integer({ minimum: 100 })` |
| `log_level` enum | union of 4 string literals |
| service `type` enum | `Union([webhook, in_process])` |
| port range 1..65535 | `Type.Integer({ minimum: 1, maximum: 65535 })` |

The `uri` FormatRegistry note is correct: TypeBox `Value.Check` ignores `format` unless registered; `URL.canParse` enforces an absolute URI (scheme required), matching ajv-formats `uri` — `not-a-url` rejected, `https://…` accepted.

### 2. Test coverage — NON-VACUOUS (verified empirically)

The 9 "validation parity" tests + `port: -1` assert the load fails specifically with `kind: "validation"`, not merely that it throws. I proved they are load-bearing by neutering `validateYamlShape` to always-succeed and re-running:

```
8 of 9 parity tests FAILED ("expected failure") with validation disabled
```

The security-critical `encryption: {}` and `masterSecret`-typo cases were among the 8 that flip to passing-load when validation is off — exactly the fail-open this guards. The lone test that still passes neutered ("unknown service type" `grpc`) is caught downstream by the Effect `YamlServiceBlock` union — a valid second-layer rejection, just not exclusively TypeBox-dependent. Acceptable.

The P3-c gap fixes are present and pass:
- `sessionCallbackTokenBothPresent` — both `timeout_ms` + `callback_token` survive the YAML decode (the projection at `YamlWebhookServiceConfig` Config.map handles the both-present branch explicitly).
- `contactsBinding` — `services.contacts: webhook` → `contactWebhook` binding, `sessionWebhook` undefined.

All 23 tests pass on a clean checkout.

### 3. No valid-path regression

`devDefaults`, `encryptionSurfaces`, `sessionsBinding`, `appsPassthrough`, `contactsBinding`, `sessionCallbackTokenBothPresent`, `envInterpolation` all load well-formed configs and assert correct projection. All 15 non-parity tests pass even with validation neutered → valid paths don't depend on rejection logic, and the schema isn't over-strict (no false rejects). The `port: -1` reframe is sound: it's now caught structurally (TypeBox `minimum: 1`) before the Effect decode and still surfaces as `kind: "validation"`; dropping the `.configError` assertion is correct since the rejection no longer originates from an Effect ConfigError.

### 4. Mechanism sound

`@sinclair/typebox` is a direct dep of `@moltzap/server` (`^0.34.0`); `ajv` + `ajv-formats` are ABSENT from the package's deps (Ajv stays removed; lockfile residue is transitive). `validateYamlShape` runs in `loadYamlFromDisk` immediately before `decodeYaml` (the lenient Effect Config decode) — mirroring `main`'s `loader.ts` running Ajv first. config.ts + config.test.ts typecheck clean (once workspace `@moltzap/protocol`/`client` are built) and lint clean.

---

### Findings

- **P3 (non-blocking, optional):** `main`'s schema.test.ts had `rejects empty database url string` (`database.url: ""`). TypeBox enforces it (`url` is `String({ minLength: 1 })`), but the new "validation parity" block has no re-asserting negative test for it. Structurally covered, just not pinned by a test. Not in the directed parity list; mention only for completeness.

No P0/P1/P2 findings. The blocking regression is fully resolved.

chughtapan:
## /codex adversarial re-review — STATUS: DONE (gate PASS, 0 findings)

**Scope:** targeted re-review of `97173aae` (`git diff 8ce2a700..97173aae`), the validation-parity fix on top of the #680 config consolidation. Confirms the previously-blocking encryption regression (Ajv→Effect-Config swap silently disabled at-rest encryption for `encryption: {}` / camelCase `masterSecret`) is now closed by the TypeBox `Value.Check` (`validateYamlShape`) running before the lenient Effect Config decode.

**Mechanism:** Codex (gpt, medium reasoning) cross-checked against the embedded diff + config context, then verified empirically against the real `@sinclair/typebox@0.34.49` schema and `ajv-formats@2.1.1` for parity. No bypass found.

### 1. Encryption (the P0 regression) — CLOSED
Every variant now FAILS loudly (empirically confirmed via `Value.Check`):
- `encryption: {}` → FAIL (`master_secret` required)
- `encryption: { masterSecret: x }` (camelCase) → FAIL (unknown key + missing required)
- `encryption: { master_secret: "" }` → FAIL (`minLength: 1`)
- `encryption: null` / non-object / array / number → FAIL
- `encryption: { master_secret: 123 }` (wrong type) → FAIL
- `encryption: { master_secret: x, extra: 1 }` → FAIL (unknown key)

No construction yields `encryptionMasterSecret: undefined` via silent relaxation. Env precedence (`ENCRYPTION_MASTER_SECRET`) is orthogonal — a malformed YAML `encryption` block fails the shape check before any decode.

### 2. Unknown-key rejection — `additionalProperties: false` at EVERY level
Confirmed rejected: top-level `bogus`, nested `server.extra`, `database.extra`, `dev_mode.extra`, `registration.extra`, `apps[].extra`, `services.<unknown-child>`, webhook-branch extra key, in_process-branch extra key. Retired top-level `seed` block → FAIL. The `Type.Union([Webhook, InProcess])` does NOT short-circuit: a key unknown to both closed branches is rejected by both.

### 3. Formats / bounds / enums — parity holds
- `webhook_url` via `FormatRegistry.Set("uri", URL.canParse)`: empirically MATCHES `ajv-formats` `uri` for all relevant cases — both accept absolute URIs incl. `javascript:`/`foo:bar`/`mailto:` (scheme present), both reject `not-a-url`, relative `/foo`, and empty string. The `javascript:`/`foo:bar` acceptance is NOT a regression; the prior Ajv schema accepted them too (`uri` = "absolute URI", not "http-only").
- `timeout_ms: 50` → FAIL (min 100); `100.5` float → FAIL (Integer). `log_level: verbose` → FAIL; `type: grpc` → FAIL.
- `port`: `-1`, `0`, `70000`, `NaN`, `Infinity` (incl. YAML `.inf`/`.nan`) all → FAIL (Integer 1..65535).

### 4. Valid path NOT broken
Empty config `{}` PASSES (all blocks optional); a fully-populated valid config PASSES. `port: -1` is now caught by TypeBox (`minimum: 1`) before the Effect `portNumber` decoder — the test's switch to `kind == "validation"` + dropping the `configError` assertion is correct, not a regression (same failure class, earlier structural layer).

### 5. Other bypass probes — all closed
`__proto__` injected as a YAML key on `encryption` → FAIL (own-enumerable extra key, rejected by `additionalProperties: false`; no prototype pollution path). `Value.Check` does not coerce (`port: "3000"` string → FAIL). Array entries (`apps[]`) validated against the closed `AppRefShape`.

**Verdict:** No P0/P1/P2/P3 findings. The silent-plaintext encryption regression is fully closed; full Ajv structural parity is restored via TypeBox with Ajv removed. GATE: PASS.

_Cross-model: codex (medium) + empirical TypeBox/ajv-formats probes agree. Read-only review; no edits, no merge._

PR #709, opened 2026-05-26
refactor(server): delete adapters/, replace WebhookClient with @effect/platform/HttpClient

chughtapan:
Closes the bespoke `WebhookClient` (282 lines) and the entire
`packages/server/src/adapters/` folder (5 files, ~700 lines). Replaces with
the standard `@effect/platform/HttpClient` Tag — already a transitive dep
(`@effect/platform@0.96.1`, `@effect/platform-node@0.106.0`).

Plan: [/home/tapanc/.claude/plans/adapters-delete.md](file:///home/tapanc/.claude/plans/adapters-delete.md)

## What replaced `WebhookClient.call(...)`

Thin Effect-native shape per call site:

```ts
httpClient.execute(
  HttpClientRequest.post(url).pipe(
    HttpClientRequest.setHeader("X-MoltZap-Event", "..."),
    HttpClientRequest.bodyUnsafeJson({...}),   // OR bodyText(payload) when HMAC-signed
  ),
).pipe(
  Effect.flatMap(HttpClientResponse.filterStatusOk),
  Effect.flatMap(HttpClientResponse.schemaBodyJson(ResponseSchema)),  // or asVoid for fire-and-forget
  Effect.timeout(Duration.millis(timeoutMs)),
  // fail-closed handler...
)
```

The standard `HttpClient.HttpClient` Tag is consumed directly — no moltzap-named
wrapper Tag. Production wiring lives in `app/server.ts → HttpClientLive`
(`NodeHttpClient.layerUndici` wrapped via `HttpClient.transform` to preserve
the concurrency cap).

## Where the 2 services landed

| From | To |
|---|---|
| `packages/server/src/adapters/webhook-contact-service.ts` | `packages/server/src/identity/services/webhook-contact-service.ts` |
| `packages/server/src/adapters/webhook-session-validator.ts` | `packages/server/src/identity/services/webhook-session-validator.ts` |

Each carries fresh focused tests (`webhook-{contact-service,session-validator}.test.ts`)
covering wire shape + every success branch + every fail-closed branch + a
`fast-check` property asserting fail-closed across the entire remote-failure
space. The old monolithic `webhook.test.ts` (293 lines) is gone.

## `signWebhookPayload` decision

**Relocated** to `packages/server/src/crypto/webhook-signature.ts`. Sole caller is
`task/services/message.service.ts → fireDeliveryWebhook` (the only HMAC-signed
outbound webhook in the repo). The helper sits next to other envelope/sig code
in `crypto/`. Behavior unchanged; the function body is identical.

HMAC byte-exactness preserved via `HttpClientRequest.bodyText(payload, "application/json")`.
`bodyText` writes the pre-serialized string verbatim; `bodyJson`/`bodyUnsafeJson`
would re-stringify the object and drift the signature off the wire bytes — an
explicit comment in `fireDeliveryWebhook` flags this for future readers.

## Concurrency-cap decision: KEPT

The prior `Effect.Semaphore(10)` is preserved by wrapping the
`NodeHttpClient.layerUndici`-provided HttpClient in `HttpClient.transform(...,
(effect) => permits.withPermits(1)(effect))` inside `HttpClientLive`. **Semantic
delta**: the semaphore now covers `httpClient.execute(request)` (resolved at
headers-arrival), where the old `WebhookClient.call(...)` held the permit
through body-read + decode + per-attempt timeout. Slightly weaker but still
bounds concurrent outbound requests; acceptable for the fire-and-forget
delivery webhook (the only burst-prone call site under CoreApp's HttpClient).

The standalone HttpClient (used by the YAML-wired contact/session validators)
does NOT carry a semaphore: it talks to a different endpoint than the delivery
webhook, and those calls are naturally per-request-bounded by inbound
concurrency. The old code had one shared semaphore across all three call
sites; this is a deliberate split.

## Auto-fix during pre-PR review: standalone Undici dispatcher scope-disposal (P1)

Caught during the `/review` pass before pushing. The first cut of
`standalone.ts` did:

```ts
const httpClient = yield* HttpClient.HttpClient.pipe(
  Effect.provide(NodeHttpClient.layerUndici),
);
```

`NodeHttpClient.layerUndici` is `Layer.scoped` over a fresh `Undici.Agent`
with an `Effect.acquireRelease(... dispatcher.destroy())` finalizer. The
`Effect.provide` scope closes the moment the yield returns the client value,
**immediately destroying the Agent**. Subsequent calls from
`WebhookSessionValidator` / `WebhookContactService` would issue against a
destroyed dispatcher. No in-tree test exercises the YAML-wired standalone
webhook path, so this would have shipped silent.

**Fix**: use `NodeHttpClient.layerUndiciWithoutDispatcher.pipe(Layer.provide(NodeHttpClient.dispatcherLayerGlobal))`.
The process-global Undici dispatcher has no per-instance lifecycle, matching
this client's server-lifetime role.

The CoreApp's `HttpClientLive` keeps `layerUndici` because it IS properly
scoped under `ManagedRuntime`, and the dispatcher disposes cleanly on
`app.close() → dispatchRuntime.dispose()`.

Reviewers: please scrutinize the standalone wiring + the explanatory comment
in `startServerEffect`.

## Confirmations

- [x] `packages/server/src/adapters/` does NOT exist (folder + all 5 files gone) — `git ls-files packages/server/src/adapters/` returns empty.
- [x] `CoreConfig.webhookClient` removed (public-surface change; the field + its `WebhookClient` import are gone from `config.ts`).
- [x] `WebhookClient`, `WebhookClientTag`, `WebhookHttpError`, `WebhookTimeoutError`, `WebhookNetworkError`, `WebhookDecodeError` all gone — `git grep` returns 0 matches.
- [x] `signWebhookPayload` lives at `packages/server/src/crypto/webhook-signature.ts`; sole caller verified by `git grep -l signWebhookPayload`.
- [x] `HttpClient` Tag wired via the standard `@effect/platform-node` layer in `app/layers.ts` + `app/server.ts` (no moltzap-named wrapper).
- [x] Behavior preserved on the 3 call sites — same wire (URL, method, headers, body bytes), same response decoding, same fail-closed semantics, same timeout.

## Gates (all green)

- `pnpm typecheck` — pass (pre-commit hook ran without `--no-verify`)
- `pnpm lint` — pass (workspace-wide, including knip + eslint + oxlint)
- `pnpm test` — 188/188 (server unit), 148/148 (server integration), 45/45 (server conformance), all other workspace tests pass
- `pnpm build` — all packages
- `./scripts/sloppy-code-guard.sh` — pass

## Acceptable notes (5)

1. **Concurrency cap delta** — semaphore covers headers-arrival, not full body-read. Bounded but weaker. Acceptable for delivery webhook bursts.
2. **Standalone HttpClient has no semaphore** — different endpoint than delivery, naturally bounded by inbound concurrency. Deliberate split.
3. **HMAC byte-exactness preserved** — explicit `bodyText(payload, "application/json")` + inline comment.
4. **Fail-closed semantics preserved + fast-check verified** — property test asserts the invariant across the full remote-failure space (arbitrary non-2xx, non-JSON, schema-mismatched, transport-failure).
5. **Retry semantics preserved** — `Effect.timeout` produces a `TimeoutException` in the error channel, so `Effect.retry` retries on timeout, matching the old `timeoutFail` behavior.

## Pre-PR gates run

- `/review` — found + auto-fixed the standalone Undici dispatcher P1 (see Auto-fix section above); 5 informational notes documented here.
- `/simplify` — not available in this harness (skill not installed); covered by the `/review` pass which evaluated structural simplifications.

---

Generated with [Claude Code](https://claude.com/claude-code)

---

## Stamina round 1 outcome

Stamina found **1 P2 + 3 P3s**. Verdict on unanimously clean dimensions:
HMAC byte-exactness, dispatcher lifecycle, TLS, headers, sensitive-data
egress, `CoreConfig` removal, error mapping — all pass.

### P2 (fixed in `aae3dbfd`): standalone validators lost the shared concurrency cap

On main, ONE `WebhookClient(10)` semaphore covered all 3 outbound webhook
paths. My first cut wrapped only the CoreApp's delivery `HttpClient` with
the cap; the standalone path (session/contact validators) was uncapped, so
validator requests could exceed 10 in-flight against an operator's
collector. The `HttpClientLive` JSDoc also literally claimed "Preserves
the `Effect.Semaphore(10)` behavior" — false for the standalone path.

**Fix (option 1, shared cap)**: extracted the semaphore into a new
`packages/server/src/app/outbound-webhook-cap.ts` module. It owns a
module-internal `Effect.Semaphore(10)` constructed once at import (matching
the deleted `new WebhookClient()` shape), and exports
`applyOutboundWebhookCap(client)`. Both `app/server.ts → HttpClientLive`
and `standalone.ts` call the same helper, so every outbound webhook in the
process pulls from the SAME permit pool — "one process, one cap" exactly
as the deleted `WebhookClient` did. JSDoc updated to match.

### P3 #1 (fixed): response bodies not drained on `filterStatusOk` failure / `asVoid`

Every webhook call now does `Effect.tap((response) => response.text)`
before `filterStatusOk`. `response.text` is `Effect.cached`, so the
subsequent `schemaBodyJson` reuses the buffer on 2xx; on non-2xx the
socket buffer is consumed immediately rather than waiting on the
FinalizationRegistry. Applied to all 3 call sites.

### P3 #2 (fixed): no regression test pinning HMAC byte-exactness

New `packages/server/src/task/services/message-service-delivery-hmac.test.ts`
constructs a request, captures what the test HttpClient sees on the wire,
and asserts `signWebhookPayload(secret, captured_body_bytes)` equals the
`X-MoltZap-Signature` header. Any future refactor that swaps
`HttpClientRequest.bodyText(payload)` for `bodyJson(obj)` /
`bodyUnsafeJson(obj)` would re-stringify the object and silently drift
the signature — this test catches it at the contract level instead of
relying on a downstream receiver to notice.

### P3 #3 (deferred — out of scope): `as AgentId` / `as UserId` casts

Pre-existing in `webhook-session-validator.ts`. Belongs in a follow-up
`Schema.fromBrand` cleanup for the identity types, called out in the
CHANGELOG under "Internal notes".

## Documentation

- **CHANGELOG.md** — new `[Unreleased]` subsection `### adapters/ deleted — webhook transport unified on @effect/platform/HttpClient (#709)` describing every change in the refactor + the stamina fix.
- Workspace-root + per-package docs: no other stale references to `WebhookClient` / `adapters/` / "WebhookClient(10)" remain (verified by `grep -rn`). The earlier `packages/server/CLAUDE.md`, `packages/server/src/identity/README.md`, and `packages/server/src/transport/README.md` updates from commit `3c8e1680` cover everything else.

## Gates (post-fix, all green)

- `pnpm typecheck` — pass (pre-commit hook ran without `--no-verify`; final hook output: `All guards passed.`)
- `pnpm lint` — pass (workspace-wide: knip + eslint + per-package + oxlint)
- `pnpm test` (server unit) — **189/189** (previously 188; +1 for the new HMAC regression test)
- `pnpm test:integration` (server) — 148/148
- `pnpm test:conformance` (server) — 45/45
- `pnpm build` — all packages
- `./scripts/sloppy-code-guard.sh` — pass

---

## Stamina round 2 outcome — deferred P3 folded in

Review-senior's "pre-existing `as AgentId` / `as UserId` casts in
session-validator" P3 was originally deferred to a follow-up. Folded
into this PR instead — the casts are now gone and three new tests pin
the fix.

### Mechanism (option B from dispatch)

Keep the existing Effect Schema for response shape; brand-decode the
id fields through the canonical TypeBox schemas via `Value.Decode`:

- `webhook-session-validator.ts` replaces the synchronous `Effect.map`
  mapper that ended in `result.agentId as AgentId` with a
  `decodeSessionValidation` helper that runs each id through
  `Value.Decode(AgentId, ...)` / `Value.Decode(UserId, ...)`. Decode
  failures raise a new typed `SessionValidationBrandDecodeFailed`
  (TaggedError) that flows through the existing `catchAllCause`
  fail-closed handler — same posture as HTTP / network / timeout /
  schema-decode errors: `{ valid: false }`.

- `@moltzap/protocol/identity` barrel now re-exports the runtime
  TypeBox schemas (`AgentId` / `UserId` / `ContactId`) — previously
  only the static types. Matches the convention
  `@moltzap/protocol/task` already uses for `AppId` /
  `ConversationId`. All existing `import type` consumers keep
  working — TypeScript's `export { X }` covers both the value and
  the static type from the schema.

Option A (`Schema.fromBrand`) was rejected — our brands are pure
TypeScript phantoms via `Brand.Brand<...>` (no Effect-native brand
constructor), so wiring `Schema.fromBrand` would require introducing
one. Option C (full TypeBox in the response schema) would have
required deleting the Effect Schema entirely. Option B is the
narrowest fix that preserves both sides of the existing structure.

### New tests (3, all pass)

- `returns invalid when agentId is empty string`
- `returns invalid when agentId is not a UUID`
- `returns invalid when ownerUserId is not a UUID`

Pre-existing success tests updated their `"agent-1"` / `"user-1"`
placeholder constants to real UUIDs — the whole point of the fix is
that those placeholders no longer leak through. Test count for
`webhook-session-validator.test.ts`: 9 → 12. Server unit total: 189
→ 192.

### Scope discipline

The 5 other `as AgentId` / `as UserId` casts elsewhere in the server
(`app-host.ts`, `agent-visibility.ts`, `presence.service.ts`,
`message.service.ts`) are intentionally left alone — they have their
own contexts and belong in a separate follow-up. This commit only
kills the 2 casts review-senior flagged on
`webhook-session-validator.ts`.

### Gates (all green)

- `pnpm typecheck` — pass (pre-commit hook ran without `--no-verify`;
  final hook output: `All guards passed.`)
- `pnpm lint` — pass (workspace-wide: knip + eslint + per-package +
  oxlint)
- `pnpm test` (server unit) — **192/192** (+3 vs round 1; the new
  fail-closed brand-decode tests)
- `pnpm test:integration` (server) — 148/148
- `pnpm build` — all packages
- `./scripts/sloppy-code-guard.sh` — pass


chughtapan (commented):
## Security review (stamina-dispatched) — PR #709

**Verdict: CLEAN.** No P0/P1/P2 findings. One P3 informational note.

Reviewed: commit `3c8e1680` on branch `impl/adapters-delete`. Reviewed against the six focus areas in the dispatch brief (HMAC byte-exactness, TLS verification, header preservation, sensitive-data egress, semaphore preservation, global-dispatcher behaviour).

### 1) HMAC byte-exactness — CORRECT

`message.service.ts → fireDeliveryWebhook` builds the wire body with:

```ts
const payload = JSON.stringify(body);                       // produce bytes ONCE
const signature = signWebhookPayload(cfg.secret, payload);   // sign those exact bytes
…
HttpClientRequest.bodyText(payload, "application/json"),    // write those exact bytes
```

`@effect/platform`'s `bodyText` (internal/httpClientRequest.js → `setBody(internalBody.text(body, contentType))`) writes the supplied string verbatim — it does not re-serialize. Verified at the dependency boundary (`@effect/platform@0.96.1`). `bodyJson`/`bodyUnsafeJson` would have re-stringified and drifted the signature; an inline comment in `fireDeliveryWebhook` also flags this for future readers. Grep across the whole repo confirms `bodyJson`/`bodyUnsafeJson` are NOT used anywhere on the signed path.

### 2) TLS verification — DEFAULT (ON)

Both transport wirings rely on Undici defaults — no TLS-disabling option:

- `app/server.ts → HttpClientLive`: `NodeHttpClient.layerUndici` — creates `new Undici.Agent()` with no options (Node defaults; `rejectUnauthorized: true`).
- `standalone.ts → startServerEffect`: `NodeHttpClient.layerUndiciWithoutDispatcher.pipe(Layer.provide(NodeHttpClient.dispatcherLayerGlobal))` — uses `Undici.getGlobalDispatcher()`, i.e. the Node process-global agent (same defaults).

Grep for `rejectUnauthorized`, `allowUnauthorized`, `insecure`, `tls` across `packages/server/src/` returns zero matches. No TLS bypass surface added by this PR.

### 3) Header preservation — CORRECT

- Delivery webhook (signed): `setHeaders({"Content-Type": "application/json", "X-MoltZap-Event": "messages.delivered", "X-MoltZap-Signature": signature})` then `bodyText(payload, "application/json")`. `bodyText`'s content-type argument re-asserts `application/json` (case-insensitive merge); both values are identical, so the final wire header is `Content-Type: application/json`. All three headers ship.
- Session validator (unsigned): `setHeader("X-MoltZap-Event", "sessions.validate")` + `bodyUnsafeJson({token})`. `bodyUnsafeJson` sets `Content-Type: application/json` via `setBody`. Both headers ship.
- Contact service (unsigned): same shape with `"contacts.check"`. Both headers ship.

### 4) Sensitive-data egress — NO REGRESSION

- Session validator body carries only `{token}`; contact service body carries only `{userIdA, userIdB}`. Nothing extra added.
- Error logging: session validator logs `{cause: Cause.pretty(cause), url}` — `cause` here is one of `HttpClientError.ResponseError | ParseError | RequestError | TimeoutException`, none of which embed the request body. Contact service logs `{err: String(err), userIdA, userIdB, url}` (operational userIds, no secrets).
- The OLD `WebhookNetworkError.cause` carried the raw `fetch`-thrown value (`err`), which in some Node failure modes can include the request URL but not the body. The NEW path's `String(err)` similarly does not surface the request body. No new egress channel; if anything, the new shape is slightly tighter because the `cause` is now an Effect tagged error rather than a raw thrown value with arbitrary fields.
- The HMAC secret is never logged or surfaced in any error path (it lives in `cfg.secret` and is consumed only by `signWebhookPayload`).

### 5) Semaphore concurrency cap — PRESERVED

`app/server.ts → HttpClientLive`:

```ts
const inner = yield* HttpClient.HttpClient;
const permits = yield* Effect.makeSemaphore(OUTBOUND_WEBHOOK_CONCURRENCY); // = 10
return HttpClient.transform(inner, (effect) => permits.withPermits(1)(effect));
```

`HttpClient.transform` wraps every `execute` call inside `permits.withPermits(1)`. The semaphore is created once per layer instantiation (singleton via `Layer.effect`), so all `MessageService.fireDeliveryWebhook` calls share the same 10-permit pool. The standalone HttpClient deliberately omits the semaphore (documented in PR body); that surface is naturally bounded by inbound concurrency. Per the dispatch brief, DoS-class concerns are not in scope for this review, but I confirm the cap claim is technically accurate.

### 6) Auto-fixed Undici `dispatcherLayerGlobal` — SAFE

`dispatcherLayerGlobal` resolves to `Undici.getGlobalDispatcher()` — Node's process-global Undici agent. Implications:

- TCP connection pool is shared across any other consumer of the global dispatcher in the same Node process. For a server process where the standalone is the sole entry point, this is a non-event.
- No new security surface: same defaults as `layerUndici` for TLS, headers, redirects.
- Not a defect — actually the correct lifecycle choice for this scope (the explanatory comment in `startServerEffect` makes the reasoning clear).

The fix replaces a real bug (the previous `Effect.provide(layerUndici)` would have destroyed the dispatcher immediately, then issued requests against a dead Agent — silently failing closed and rejecting all sessions). The fix is sound.

---

### P3 (informational, non-blocking)

**HMAC signature has no in-tree regression test.** No test asserts that `fireDeliveryWebhook` produces a `X-MoltZap-Signature` whose hex matches `HMAC-SHA256(secret, wireBody)`. The pattern in source is correct and the inline comment is load-bearing for maintainers, but if a future refactor swaps `bodyText` for `bodyJson` the regression would ship silent. Consider adding a focused unit test that captures the outbound `HttpClientRequest`, extracts the body via the same helper used in `webhook-session-validator.test.ts → extractBodyText`, and recomputes the HMAC for equality. Not a blocker — the prior `webhook.test.ts` deleted by this PR didn't cover this either.

### Excluded by review scope

- Integration tests under `packages/server/src/__tests__/` are excluded from `tsc --build`; per dispatch brief, not in scope.
- Standalone HttpClient lacking a semaphore is DoS-class only — excluded per review charter.

---

VERDICT: DONE (clean)

chughtapan:
## Verdict
APPROVE-PENDING-CI

**Role:** acceptance-vs-diff stamina pass (single-reviewer in this dispatch; orchestrator coordinates stamina N).
**Branch/commit:** `impl/adapters-delete` @ `3c8e168`.
**Aggregate rule:** all checks pass on diff-static review; CI in progress at time of review — APPROVE conditioned on CI green.

### Acceptance criteria — every plan AC verified against the diff

| AC | Verdict | Evidence |
|---|---|---|
| 1. `adapters/` folder + 5 files gone | PASS | `ls packages/server/src/adapters/` → does-not-exist; `git ls-files` empty. |
| 2. `WebhookClient` + 4 tagged errors + `WebhookClientTag` gone | PASS | `git grep` returns 0 matches in code; one residual JSDoc reference in `app/server.ts → HttpClientLive` doc-comment is historical-context narration (acceptable per cold-start-reader: explains why semaphore exists). |
| 3. 2 webhook services in `identity/services/` use `HttpClient` directly | PASS | `webhook-contact-service.ts` + `webhook-session-validator.ts` accept `HttpClient.HttpClient` via constructor; use `.execute(HttpClientRequest.post(...).pipe(...)).pipe(filterStatusOk, schemaBodyJson, timeout, catchAll/catchAllCause)`. |
| 4. `message.service.ts` delivery webhook uses `HttpClient`, `webhookClient` dep dropped | PASS | `fireDeliveryWebhook` uses `this.httpClient.execute(...)`; `MessageServiceDeps.httpClient: HttpClient.HttpClient`; no `webhookClient` field. |
| 5. `CoreConfig.webhookClient` removed | PASS | `git grep -n 'webhookClient' -- packages/` returns 0 matches. |
| 6. `HttpClient` Tag wired in `app/layers.ts` via `@effect/platform-node` | PASS | `DeliveryWebhookTag` doc cites `NodeHttpClient.layerUndici`; `MessageServiceLive` consumes `HttpClient.HttpClient` from context. |
| 7. HMAC byte-exactness preserved | PASS | `fireDeliveryWebhook`: `const payload = JSON.stringify(body); const signature = signWebhookPayload(cfg.secret, payload); ... HttpClientRequest.bodyText(payload, "application/json")` — same string signed and sent. NOT `bodyJson/bodyUnsafeJson`. Inline comment flags the pitfall. |
| 8. Build + lint + test green | PASS LOCAL | `tsc --noEmit` on `@moltzap/server-core` → exit 0; `pnpm -w lint` → 0 warnings/0 errors across 544 files; 188/188 server unit tests green including 8 contact + 9 session new tests. CI still in progress. |
| 9. Arch docs updated; pre-commit hooks pass | PASS | `packages/server/CLAUDE.md` removes `adapters/` row + adds webhook-HMAC mention to `crypto/`; `identity/README.md` lists the two new webhook services; `transport/README.md` mermaid swaps `WebhookClient` → `HttpClient`; consumer JSDoc in `contact-policy.ts` + `session-validator.ts` re-points to new location. |

### Behavior preservation on the 3 call sites

- **`WebhookContactService.areInContact`** — same POST URL + `X-MoltZap-Event: contacts.check` + JSON body `{userIdA, userIdB}`; decodes `{inContact: boolean}`; `catchAll(...).as(false)` fail-closed; `Effect.timeout(timeoutMs)` preserved. Wire-shape test asserts URL+method+event+body. Property test (`fast-check`, 64 runs) asserts ALWAYS returns `false` / NEVER fails across the entire universe of remote failures (arbitrary non-2xx, non-JSON, schema-mismatch, transport-fail).
- **`WebhookSessionValidator.validateSession`** — same POST URL + `X-MoltZap-Event: sessions.validate` + body `{token}`; decodes the discriminated `{valid: true, agentId, ownerUserId, agentStatus?} | {valid: false}` union; `catchAllCause` fail-closes; defects logged at `error`, expected failures at `warning`. Property test mirrors contact-service shape.
- **`message.service.ts → fireDeliveryWebhook`** — same POST URL + headers (`Content-Type`, `X-MoltZap-Event: messages.delivered`, `X-MoltZap-Signature`); body via `bodyText` (HMAC byte-exact); `filterStatusOk + timeout + retry(exponential + jittered, capped at MAX_ATTEMPTS-1) + asVoid + catchAll(logError)` preserves fire-and-forget semantics + retry semantics (`Effect.timeout` raises a `TimeoutException` in the error channel, which `Effect.retry` retries on, matching the prior `timeoutFail` behavior). No in-tree test exercises this path (acknowledged in PR body); behavior-preservation argument rests on shape inspection + `Effect.retry` doctrine.

### Undici dispatcher lifecycle — auto-fix verified sound

The standalone wiring uses `NodeHttpClient.layerUndiciWithoutDispatcher.pipe(Layer.provide(NodeHttpClient.dispatcherLayerGlobal))`. Verified against `@effect/platform-node@0.106.0` source: `dispatcherLayerGlobal = Layer.sync(Dispatcher, () => Undici.getGlobalDispatcher())` is non-scoped (no finalizer); `layerWithoutDispatcher = Client.layerMergedContext(Effect.map(Dispatcher, make))` is non-scoped (constructs a pure value, no acquireRelease). The original bug was that `layerUndici = Layer.provide(layerWithoutDispatcher, dispatcherLayer)` carries the scoped `dispatcherLayer = Layer.scoped(Dispatcher, makeDispatcher)` finalizer (`Effect.acquireRelease(... dispatcher.destroy())`) that fires when the surrounding `Effect.provide` closes. The fix dodges this by substituting the global dispatcher.

CoreApp's `HttpClientLive` keeps `layerUndici` because it sits under `ManagedRuntime`, so the scoped finalizer is coterminous with `dispatchRuntime.dispose()` (called by `app.close()`). No leak; no early destroy.

### Concurrency cap delta — disclosed, judged acceptable

`HttpClient.transform(inner, (effect) => permits.withPermits(1)(effect))` wraps the `Effect<HttpClientResponse, ...>` that resolves at headers-arrival — `transform`'s `f` parameter operates on the pre-decoded response effect. The prior `WebhookClient.call(...)` held the permit through the full body-read + decode + per-attempt timeout. For fire-and-forget delivery webhooks with small bodies and `Effect.timeout(DELIVERY_WEBHOOK_TIMEOUT_MS)` clipping the post-headers tail, the new semantics still bound burst concurrency at headers; weaker but acceptable. If hostile remote-server hangs during body-read become an operational concern, a follow-up can wrap the body pipeline inside the semaphore.

### Test-override path — confirmed

`packages/server/src/app/layers.test.ts` uses `FetchHttpClient.layer` in `BaseLive` to satisfy the `HttpClient.HttpClient` requirement. Standard test pattern; the unit tests for the two webhook services use `Layer.succeed(HttpClient.HttpClient, fakeClient)` via direct constructor injection (the services accept `HttpClient.HttpClient` as a constructor arg). Both patterns work.

### P1-P4 craft floor

- **P1 (Types beat tests).** Two pre-existing `as AgentId` / `as UserId` casts at the SessionValidator boundary (`webhook-session-validator.ts → validateSession`, the result-mapping `result.agentId as AgentId`). Pre-existed in `adapters/webhook-session-validator.ts:41-42`; this PR is a verbatim relocation. Schema could use `Schema.fromBrand`. **P3 — non-blocking; out of scope as pre-existing.**
- **P2 (Validate at boundaries).** Every HTTP ingress decodes through `HttpClientResponse.schemaBodyJson(Schema)`. Clean.
- **P3 (Errors are typed).** No `throw new Error`, no silent catch, no `catch (e) { return null }`. `Effect.catchAll(...).as(false)` and `Effect.catchAllCause(... { valid: false })` are typed error-channel collapses with explicit logging. Clean.
- **P4 (Exhaustiveness).** `result.valid !== true` branch handles the discriminant; the cause-handler in session-validator uses `Cause.dieOption(cause)._tag === "Some"` to split defect-vs-failure logging — both branches explicit. Clean.

### Local gates

- `pnpm --filter @moltzap/server-core exec tsc --noEmit` → exit 0
- `pnpm -w lint` → 0 warnings, 0 errors, 544 files
- `pnpm --filter @moltzap/server-core test -- --run src/identity/services/webhook-{contact-service,session-validator}.test.ts` → 17/17 passed
- Full `pnpm --filter @moltzap/server-core test` → 188/188 server unit tests passed
- Knip not installed locally; configured for CI; trusting CI gate (`ci` job currently `IN_PROGRESS`).

### CI status
3 checks IN_PROGRESS at review time (`ci`, `conformance`, `test`); `Mintlify Deployment` SKIPPED. **CI must report green before merge.** Per Phase 1a CI-gate, this caps the verdict at `APPROVE-PENDING-CI` — the orchestrator's auto-monitor must withhold merge until the rollup clears.

### Process issues
none.

chughtapan:
## Stamina review: codex (adversarial), medium reasoning

**Status: DONE_WITH_CONCERNS** — 1 P2 (real behavior change) + 1 P3 (advisory) confirmed against the diff. No P0/P1 ship-blockers found.

Cross-checked codex's 4 findings against `gh pr diff 709` and the full tree at `origin/impl/adapters-delete`. Dismissed 2 phantoms.

---

### Confirmed findings

**[P2] Standalone session/contact webhooks lost the shared concurrency cap.** *(codex P1; lowered to P2 — real, but bounded)*

On `main`, `standalone.ts` constructed ONE `WebhookClient` (carrying its own `Effect.Semaphore(10)`) and passed it to:
1. Core delivery webhook (via `config.webhookClient`)
2. `WebhookSessionValidator` (YAML-wired)
3. `WebhookContactService` (YAML-wired)

All three call sites SHARED that single semaphore — a global outbound-webhook cap.

In the PR (`standalone.ts → startServerEffect`):
- Lines 293-299 build a raw `HttpClient` from `layerUndiciWithoutDispatcher + dispatcherLayerGlobal` with **no semaphore**, passed to both validators.
- `createCoreApp(coreConfig)` then builds its OWN `HttpClientLive` (`app/server.ts → HttpClientLive`) with a fresh `Effect.Semaphore(10)` for delivery only.

Net behavior change in standalone deployments:
- Delivery webhook: capped at 10 (preserved, but uses a separate dispatcher pool).
- Session/contact validators: **uncapped**.
- Across-call-site shared cap: **gone** — delivery and validators no longer contend on one budget.

The JSDoc in `app/server.ts → HttpClientLive` claims "Preserves the `Effect.Semaphore(10)` behavior of the prior bespoke `WebhookClient`" — this is FALSE for standalone session/contact paths. Plan §Guardrails line 58 said "decide explicitly whether the `Effect.Semaphore(10)` is doctrine (mention it in the PR body either way)" — the decision wasn't documented in the PR body.

Fix options: (a) wire the same cap on the standalone httpClient via `HttpClient.transform` before passing to validators; (b) reuse the core's `HttpClient.HttpClient` Tag from inside the ManagedRuntime instead of building a separate one; (c) explicitly document the deliberate behavior change in the PR body + JSDoc.

**[P3] Response bodies not drained on success-discard and non-2xx paths.** *(codex P1; lowered to P3 — bounded by FinalizationRegistry)*

- `fireDeliveryWebhook` (`message.service.ts → fireDeliveryWebhook`) does `execute → filterStatusOk → asVoid` — body never read.
- `webhook-session-validator.ts → validateSession` and `webhook-contact-service.ts → areInContact` read body via `schemaBodyJson` on the 2xx path but `filterStatusOk` fails before `schemaBodyJson` on non-2xx, leaving the Undici body buffered.

On `main`, `WebhookClient.call` always called `response.text()` (see `parseWebhookResponse → readResponseText` in deleted `adapters/webhook.ts`), draining the body unconditionally.

In `@effect/platform/internal/httpClient.js`, `responseRegistry.register(response, controller)` is a `FinalizationRegistry` that aborts the controller (closing the Undici stream) on response GC. So leaks are eventual, not permanent — but GC timing is unpredictable under burst load. With Undici's keep-alive, undrained bodies hold connection slots until the GC fires.

Real-world impact: low. Delivery webhook is fire-per-message-with-offline-recipients (low volume); validators are per-auth-request. For the typical operator endpoint returning a small JSON body or 204, this won't bite. Worth fixing-in-place by calling `response.text` (and discarding) or `Stream.runDrain` on the response stream, but not ship-blocking.

---

### Dismissed phantoms

**[P2 → dismissed] Close ordering allows new fibers after `messageService.close()`.** *(codex P2)*

Codex correctly identifies the structural race: `closeCoreAppEffect` snapshots `deliveryWebhookFibers`, interrupts them, then sleeps `SHUTDOWN_DRAIN_MS` while in-flight RPCs may still call `spawnDeliveryWebhook`. New fibers added after the snapshot escape interruption.

**But this is identical on `main`** — verified `git show origin/main:packages/server/src/task/services/message.service.ts → MessageService.close` has the same `[...this.deliveryWebhookFibers]` snapshot + `clear()` pattern. This PR does not touch `spawnDeliveryWebhook`'s tracking logic. Pre-existing concern, out of scope for an "adapters delete" transport-swap PR.

**[P2 → dismissed] Server semaphore loses old "request + body" permit semantics.** *(codex P2)*

Codex argues `HttpClient.transform(inner, e => permits.withPermits(1)(e))` releases the permit at headers-arrival, before body reads. True architecturally.

But the only consumer of `HttpClientLive` (the semaphore'd client) is `fireDeliveryWebhook` (via Tag injection in `MessageServiceLive`). That call site discards the body with `asVoid` — there IS no body-read phase to protect. The two validators bypass `HttpClientLive` entirely (they use the standalone-built httpClient), so the semantic delta on the wrapped client is moot.

This finding collapses into the confirmed P2 above (standalone validators have no cap at all).

---

### Cross-checked: no regression on the brief's other concerns

1. **Dispatcher lifecycle:** `NodeHttpClient.layerUndici = Layer.provide(layerWithoutDispatcher, dispatcherLayer)` where `dispatcherLayer = Layer.scoped(Dispatcher, acquireRelease(new Undici.Agent(), d => d.destroy()))`. Wired into the `ManagedRuntime`, so dispatcher lives for server lifetime, destroyed once on `dispatchRuntime.dispose()`. No double-close. Close path runs `messageService.close()` (interrupts pending webhook fibers) BEFORE `dispatchRuntime.dispose()` — pending fibers can't outlive the dispatcher except via the pre-existing race noted above. Standalone uses `dispatcherLayerGlobal` (`Layer.sync`, no finalizer over `Undici.getGlobalDispatcher()`) — safe inside transient `Effect.provide` scope.

2. **HMAC byte-exactness:** Confirmed clean. `fireDeliveryWebhook` pre-serializes via `JSON.stringify(body)`, passes the SAME string to both `signWebhookPayload(cfg.secret, payload)` and `HttpClientRequest.bodyText(payload, "application/json")`. Validators use `bodyUnsafeJson({token})` which internally does `JSON.stringify(body)` — byte-equivalent to the deleted `WebhookClient`'s text-mode body.

3. **Error mapping:** `grep -rnE "WebhookHttpError|WebhookTimeoutError|WebhookNetworkError|WebhookDecodeError|WebhookClient[^a-zA-Z]"` across `packages/` returns only one historical comment in `app/server.ts:55` and CHANGELOG entries. Zero `catchTag`/`Match.tag` consumers depended on the deleted tag names.

4. **`config.webhookClient` removal:** `grep -rn "webhookClient" --include='*.ts'` returns zero matches outside the deleted/edited files. No silent-ignore risk.

---

### Verdict

**SHIP after P2 (cap loss) is addressed** — either fix the cap in standalone, reuse the core HttpClient, or document the deliberate change in the PR body and the `HttpClientLive` JSDoc.

P3 body-drain is fix-in-place okay.

Recommendation: Wire `HttpClient.transform` semaphore-wrap on the standalone httpClient as well, OR delete the JSDoc claim of "Preserves the `Effect.Semaphore(10)` behavior" since standalone validators no longer have any cap. The latter is a one-line fix that makes the behavior change explicit without re-architecting.

---

*Adversarial review: codex CLI 0.132.0, `gpt-5.5`, `model_reasoning_effort=medium`, ~17s wall.*

chughtapan:
## Stamina verdict: ESCALATED — one P2 (standalone validators lost concurrency cap). HMAC + dispatcher + everything else clean.

**Mode:** pr | **N:** 3 (tier=senior, exports=2) | **CI:** green (ci+conformance+test).

### Dispatch set

| Role | Skill | Verdict | Comment |
|---|---|---|---|
| acceptance-vs-diff | `/safer:review-senior` | DONE_WITH_CONCERNS (P3) | https://github.com/chughtapan/moltzap/pull/709#issuecomment-4539500336 |
| adversarial / cross-model | `/codex` | DONE_WITH_CONCERNS (P2 blocking) | https://github.com/chughtapan/moltzap/pull/709#issuecomment-4539514725 |
| security | `/security-review` | DONE (P3) | https://github.com/chughtapan/moltzap/pull/709#pullrequestreview-4360079219 |

### Unanimously verified clean
HMAC byte-exactness preserved (`bodyText(payload)` written verbatim; same string fed to `signWebhookPayload`; verified at the `@effect/platform` dependency boundary). Undici dispatcher fix sound (verified by reading `@effect/platform-node@0.106.0` source — `dispatcherLayerGlobal` has no finalizer; CoreApp keeps scoped `layerUndici` under `ManagedRuntime`). TLS verification ON. All 3 headers flow. No sensitive-data egress. `CoreConfig.webhookClient` cleanly removed. `adapters/` folder fully gone. The 4 deleted tagged errors have zero `catchTag`/`Match.tag` downstream consumers. Test override path works via `Layer.succeed(HttpClient.HttpClient, fakeClient)`.

### Blocking finding (must fix + re-review)

**[`/codex` P2] Standalone validators lost the `Effect.Semaphore(10)` cap.** On main, one shared `WebhookClient(10)` semaphore covered all 3 outbound paths (delivery + session-validator + contact-check). This PR wraps only the CoreApp delivery path via `HttpClient.transform` in `HttpClientLive`; the standalone path constructs a raw uncapped `HttpClient` for `WebhookSessionValidator` + `WebhookContactService`. Two consequences: (a) **behavioral regression** — operator validator/contact-check requests can now exceed 10 in-flight, removing the DoS guard on operator collectors; (b) **doc lie** — `HttpClientLive` JSDoc claims "Preserves the `Effect.Semaphore(10)` behavior" but it's false for standalone validators. Fix: wrap the standalone `httpClient` with the same semaphore (or share the cap across all 3 paths), AND correct the JSDoc.

### Concerns (P3 — resolve in-place, no re-review)
- [`/codex`]: response bodies not drained on `filterStatusOk` failures + `asVoid` paths. FinalizationRegistry-bounded (eventual cleanup, not permanent leak), but worth explicit drain.
- [`/security-review`]: no regression test pinning HMAC signature byte-exactness — would catch a future refactor that swaps `bodyText`→`bodyJson` silently.
- [`/safer:review-senior`]: pre-existing `as AgentId`/`as UserId` casts in session-validator (out of scope; follow-up via `Schema.fromBrand`).

### Next route
P2 fix + targeted re-review (codex) → P3 fixes in place → `/document-release` → merge.

chughtapan:
## Verdict (round 2, targeted re-review at `aae3dbfd`)

**APPROVE-PENDING-CI** — all 7 acceptance checks pass against the diff `3c8e1680..aae3dbfd`; verdict caps at *pending CI* per `/safer:review-senior` Phase 1a (CI checks `ci`, `conformance`, `test` are `IN_PROGRESS` at this writing).

Reviewer role: acceptance-vs-diff (no full re-stamina; round-1 verified-clean dimensions are not re-litigated, only checked for regression).

### Round-1 P2 fix verification — shared `Effect.Semaphore(10)`

- `packages/server/src/app/outbound-webhook-cap.ts → OUTBOUND_WEBHOOK_PERMITS` is a module-internal `Effect.Semaphore(10)` constructed once at import time (`Effect.runSync(Effect.makeSemaphore(10))`). Single source of truth for the cap.
- `applyOutboundWebhookCap(client)` wraps an arbitrary `HttpClient.HttpClient` with that permit pool via `HttpClient.transform`. Permits return on fiber interrupt (inner request runs inside `withPermits` scope).
- Both wiring sites pull from the same instance:
  - CoreApp: `app/server.ts → HttpClientLive` calls `applyOutboundWebhookCap(inner)` (delivery webhook from `MessageService.fireDeliveryWebhook`).
  - Standalone: `standalone.ts → startServerEffect` calls `applyOutboundWebhookCap(rawHttpClient)` (session/contact validators).
- Net: all 3 outbound webhook paths (delivery + contact-check + session-validate) share **one process-wide 10-permit pool**, matching the deleted `new WebhookClient(10)` semantics. P2 closed.

### JSDoc accuracy

- `HttpClientLive` JSDoc now reads: *"wrapped with the process-wide outbound-webhook semaphore from `applyOutboundWebhookCap`. The cap is SHARED with the standalone validator wiring in `standalone.ts` so all 3 outbound webhook paths (delivery + contacts + sessions) pull from the same permit pool"*. Matches the implementation. Not aspirational.
- Standalone wiring comment was rewritten into a two-numbered-concern block; concern #2 explicitly documents the shared-pool intent.

### P3 #1 (body drain) — all 3 sites covered

- `task/services/message.service.ts → fireDeliveryWebhook` (delivery): `Effect.tap((response) => response.text)` inserted before `Effect.flatMap(HttpClientResponse.filterStatusOk)`.
- `identity/services/webhook-contact-service.ts`: same insertion before `filterStatusOk` and `schemaBodyJson(ContactsCheckResponse)`.
- `identity/services/webhook-session-validator.ts`: same insertion before `filterStatusOk` and `schemaBodyJson(SessionValidateResponse)`.

Semantic check: `response.text` is `Effect<string, E>` on `HttpIncomingMessage` (confirmed against `@effect/platform/dist/dts/HttpIncomingMessage.d.ts → readonly text: Effect.Effect<string, E>`). `Effect.tap(fn)` evaluates the inner effect; the body IS consumed (not just referenced). Because `response.text` is `Effect.cached`, the downstream `schemaBodyJson` on the 2xx path reuses the same buffer — no double-read regression. Non-vacuous.

### P3 #2 (HMAC regression test) — non-vacuous, asserts wire-bytes ≡ signed-bytes

`packages/server/src/task/services/message-service-delivery-hmac.test.ts → hmacBytesMatchWireBytes` (new, 197 LOC):

- `captureFakeClient` builds a fake `HttpClient.HttpClient` whose `execute` reads `request.body` via `extractBodyBytes` (covers `Raw` and `Uint8Array` body tags) and stashes the exact wire bytes plus the `x-moltzap-signature` and `x-moltzap-event` headers into a `CapturedRequest`.
- INVARIANT 1: `expect(recorded.signatureHeader).toBe(signWebhookPayload(DELIVERY_SECRET, recorded.bodyBytes))` — recomputes HMAC over the captured wire bytes and asserts header equality. This is the byte-exact wire ≡ signed assertion; NOT a paraphrase.
- INVARIANT 2: pins exact `JSON.stringify({conversationId, messageId, offlineRecipientAgentIds: [AGENT_B]})` — would fail if a future refactor swapped `HttpClientRequest.bodyText(payload)` for `bodyJson(obj)` and produced different bytes (different key order or whitespace).
- INVARIANT 3: pins `x-moltzap-event === "messages.delivered"` (wire-name pin).
- Uses `effectIt.live` (real clock) for the fire-and-forget fiber spawned via `Effect.runFork` inside `spawnDeliveryWebhooks`; polls up to 20 × 5ms for the side-effect.

The test would catch the exact regression class round-1 codex flagged. Confirms P3 #2.

### P3 #3 (deferred) — called out

PR body has a `### P3 #3 (deferred — out of scope): \`as AgentId\` / \`as UserId\` casts` section explicitly naming the follow-up as a `Schema.fromBrand` cleanup for identity types, with the deferral also noted in the CHANGELOG under "Internal notes". Reviewers will see this.

### Regression check (round-1 verified-clean items)

- HMAC byte-exactness in implementation (`webhook-signature.ts` + `HttpClientRequest.bodyText(payload, "application/json")` in `fireDeliveryWebhook`): **untouched** by this commit (`git diff` shows only the `Effect.tap` insertion in `message.service.ts`; no edit to the `bodyText` line or the crypto helper).
- Dispatcher lifecycle (`standalone.ts` uses `layerUndiciWithoutDispatcher` + `dispatcherLayerGlobal`): **preserved** (still lines 305-306).
- `adapters/` folder: still gone (`git ls-tree -r aae3dbfd packages/server/src/adapters/` empty).
- `CoreConfig.webhookClient`: still gone (no grep match in `config.ts`).
- TLS / headers / `WebhookClient` Tag removal: not touched.

### Final state

- 2 new files added in this commit, exactly as scoped:
  - `packages/server/src/app/outbound-webhook-cap.ts` (62 LOC, cap helper)
  - `packages/server/src/task/services/message-service-delivery-hmac.test.ts` (197 LOC, HMAC regression test)
- 5 surgical modifications: `CHANGELOG.md`, `app/server.ts` (cap extraction + JSDoc), `standalone.ts` (wiring + comment), 3× webhook `.ts` (body-drain tap), `message.service.ts` (body-drain tap + comment swap).

### CI status

`CI status: pending` — 3 checks IN_PROGRESS (`ci`, `conformance`, `test`); 1 SKIPPED (`Mintlify Deployment`). Per Phase 1a, this caps the verdict at `APPROVE-PENDING-CI`. The orchestrator's auto-monitor MUST withhold merge until CI clears green. If any check transitions to FAILURE, this verdict downgrades to `HOLD` and routes to `/safer:diagnose`.

### Aggregate

| Dimension | Verdict |
|---|---|
| P2 fix (shared semaphore across 3 paths) | APPROVE |
| JSDoc accuracy | APPROVE |
| P3 #1 (body drain — non-vacuous tap) | APPROVE |
| P3 #2 (HMAC test — non-vacuous wire-bytes assertion) | APPROVE |
| P3 #3 (deferral callout in PR body) | APPROVE |
| Regression on round-1 clean items | APPROVE (no regression) |
| Final state (only 2 new files, scope held) | APPROVE |
| CI | PENDING |

**Aggregate verdict: APPROVE-PENDING-CI.**

### Confidence

HIGH on craft (read the diff end-to-end; cross-checked `Effect.tap`/`response.text` semantics against `@effect/platform` `.d.ts`; verified HMAC test is non-vacuous by reading invariants 1-3). MED on CI (still in flight; no evidence of failure but no evidence of success either).

### Effort

(human: ~30 min / CC: ~6 min). Read-only review pass, no implementation.

### Process issues

- `/safer:review-senior` Phase 2 normally dispatches `/review` + `/simplify` + `/codex review` for PR rows. This is a *targeted* re-review (round-1 stamina already burned the heavyweight passes); per the orchestrator's brief, scope is acceptance-vs-diff only. Falling through to fallback mode would re-dispatch all three on a 403-LOC diff for a 1-P2-fix-and-3-P3-fixes confirmation — disproportionate. The brief is the explicit consumer contract here, so this verdict comes from a single acceptance-vs-diff pass rather than the full table-row dispatch. Surface this so the orchestrator can confirm the round-2 stamina policy explicitly going forward.

chughtapan:
## Targeted P2 re-review — codex (cross-model adversarial), medium reasoning

**SHA reviewed:** `aae3dbfd` (P2-fix layered on `3c8e1680`)
**Scope:** Confirm the blocking P2 — shared `Effect.Semaphore(10)` cap lost across the 3 outbound webhook paths in the original `adapters/` deletion — is now closed by the new `outbound-webhook-cap.ts` module.

### Verdict

**STATUS: DONE_WITH_CONCERNS** — P2 is RESOLVED. One [P3] test-coverage observation, non-blocking.

### Cross-model verification (codex)

> P2 status: RESOLVED. The implementation uses one module-level semaphore in `outbound-webhook-cap.ts`, and every `applyOutboundWebhookCap` call closes over that same `OUTBOUND_WEBHOOK_PERMITS` instance. In normal repo execution, both imports resolve to the same source/dist module path; I don't see a `.ts`/`.js` split or package subpath pattern that would create duplicate module instances inside one process.
>
> All 3 intended paths wrap: CoreApp `HttpClientLive` wraps at `server.ts:68`, standalone constructs one capped client at `standalone.ts:310`, then passes it to `WebhookSessionValidator` at `standalone.ts:362` and `WebhookContactService` at `standalone.ts:413`. I found no 4th production JSON webhook `httpClient.execute` path bypassing the cap.
>
> The body drain also looks correct: `response.text` is an `Effect`, and the `Effect.tap((response) => response.text)` runs before `filterStatusOk` in session, contact, and delivery paths. The `HttpClientLive` JSDoc matches the implementation, including the narrower headers-arrival scope.

### Findings tagged

**[P3] — non-blocking, test-coverage gap.** `task/services/message-service-delivery-hmac.test.ts → captureFakeClient` builds a raw fake `HttpClient` and `buildDeliveryHarness` injects it directly into `MessageService`. There's no `applyOutboundWebhookCap` import and no held-permit setup. The test protects HMAC byte exactness as named, but it would still pass if the cap were removed from production wiring entirely.

This does NOT invalidate the P2 fix — production wiring at `app/server.ts → HttpClientLive` and `standalone.ts` is what cap-wraps real delivery/session/contact paths — but a "cap-behavior" test (e.g., assert second delivery waits behind held permits) would lock the contract end-to-end. Optional follow-up, not a merge blocker.

### Break-attempts that did NOT find issues

| # | Break attempt | Result |
|---|---|---|
| 1 | Module-internal semaphore re-built per call site | One `OUTBOUND_WEBHOOK_PERMITS`, closed over by every `applyOutboundWebhookCap` call. |
| 2a | CoreApp `HttpClientLive` skips wrap | `server.ts → HttpClientLive` calls `applyOutboundWebhookCap(inner)`. |
| 2b | Standalone `WebhookSessionValidator` skips wrap | `standalone.ts:310` wraps once, passed to `makeSessionValidator(bootPlan, httpClient)`. |
| 2c | Standalone `WebhookContactService` skips wrap | Same `httpClient` flows to `installContactService(app, bootPlan, httpClient)`. |
| 3 | 4th outbound webhook path bypasses cap | None found (`httpClient.execute` callers: delivery, session, contact only). |
| 4 | `Effect.tap((response) => response.text)` — property access vs Effect execution | `response.text` is `Effect.cached`, returned as Effect, runs before `filterStatusOk` on all 3 sites. |
| 6 | JSDoc/code drift on `HttpClientLive` | JSDoc accurately describes headers-arrival scope + shared semaphore. |

### Recommendation

**Recommendation:** Merge as-is because the blocking P2 is fully closed across all 3 production webhook paths via a single shared `Effect.Semaphore(10)`, and the only remaining finding is a test that names HMAC exactness (which it does cover) and incidentally doesn't double-cover the cap — a cosmetic coverage gap, not a contract regression.

---
🤖 Cross-model verification by [codex-cli 0.132.0](https://github.com/openai/codex), invoked via `/codex` skill, medium reasoning.

chughtapan:
## Targeted re-review — brand-decode cleanup (commit 570d8022)

**Scope:** confirm the deferred P3 (`as AgentId`/`as UserId` casts at `webhook-session-validator.ts → WebhookSessionValidator.validateSession`) is correctly resolved in this PR before merge. Read-only review; no source edits.

**Verdict: CLEAN — ship.**

All 6 verification items pass. No P0/P1/P2/P3 findings.

---

### 1. Both casts gone — PASS

`git grep 'as AgentId\|as UserId' packages/server/src/identity/services/webhook-session-validator.ts` returns one match only, and it's the intentional JSDoc anti-example:

```
35: * `{ valid: false }` rather than `{ valid: true, agentId: "" as AgentId }`.
```

Inside `WebhookSessionValidator.validateSession` itself, zero casts. The old `Effect.map((result) => ... as AgentId)` mapper is replaced by `Effect.flatMap(decodeSessionValidation)`. Extraction is a clean win on the 50-line method ceiling too.

### 2. Brand is real at runtime — PASS

`decodeSessionValidation` actually calls `Value.Decode(AgentId, result.agentId)` and `Value.Decode(UserId, result.ownerUserId)` inside `Effect.try`. The `AgentId` / `UserId` schemas are `brandedId("AgentId")` / `brandedId("UserId")` in `packages/protocol/src/identity/agents.ts → AgentId, UserId`, which expand via `brandedString` to `Type.String({ format: "uuid", ... })` (see `packages/protocol/src/schema-primitives.ts → brandedId, brandedString`).

The `"uuid"` format is registered at module load:

```
packages/protocol/src/schema-primitives.ts:17  FormatRegistry.Set("uuid", (value) => typeof value === "string" && UUID_RE.test(value))
```

`Value.Decode` runs registered format checks, so `Value.Decode(AgentId, "")` and `Value.Decode(AgentId, "not-a-uuid")` throw, the `Effect.try` catches the throw and emits `SessionValidationBrandDecodeFailed`. Brand is NOT a TS phantom.

### 3. Fail-closed posture preserved — PASS

`SessionValidationBrandDecodeFailed extends Data.TaggedError("SessionValidationBrandDecodeFailed")` flows through the existing `Effect.catchAllCause` in `validateSession` (no new error channel, no new escape route). Resolved value is `{ valid: false } satisfies SessionValidation`. The cause-logger correctly classifies the brand-decode failure as a warning (not a defect — `Effect.try`'s catch handler returns an Effect failure, not a die), so the existing `Cause.dieOption` branch path is correct.

Three new tests pin the contract and are **non-vacuous** — every one asserts the RESOLVED value equals `FAIL_CLOSED` (`{ valid: false }`), not just "an error throws":

- `failsClosedOnEmptyAgentId` → `agentId: ""` → `expect(yield* validator.validateSession(TOKEN)).toEqual(FAIL_CLOSED)`
- `failsClosedOnNonUuidAgentId` → `agentId: "not-a-uuid"` → same assertion
- `failsClosedOnNonUuidOwnerUserId` → `ownerUserId: "also-not-a-uuid"` → same assertion

Existing success-branch tests correctly migrated `"agent-1"` / `"user-1"` placeholders to real UUIDs (`00000000-0000-4000-8000-0000000000a1` / `...0000000000b2`) — without that migration they'd start failing for the right reason, which would itself prove the brand is enforced. The existing property-based `remoteFailureArb` is unchanged and still covers transport/schema-failure space at `numRuns: 64`. Test count: 9 → 12 as claimed.

### 4. Supporting change is safe — PASS

`packages/protocol/src/identity/index.ts` flipped `export type { AgentId, ContactId, UserId }` → `export { AgentId, ContactId, UserId }`. All ~30 in-repo consumers use `import type { ... } from "@moltzap/protocol/identity"` (grepped); TypeScript elides those at type level even when the underlying export is a value, so zero consumer breakage. Only the new validator uses `import { AgentId, UserId }` (value).

Convention-match claim: `packages/protocol/src/task/index.ts` exports `ConversationId, LeaseId, MessageId, TaskId` (from `methods.js`) and `AppId` (from `methods.js`) all as runtime values, not `export type`. Verified — the identity barrel now matches the task barrel.

The shared dual-name idiom (`export const AgentId = brandedId("AgentId"); export type AgentId = Static<typeof AgentId>;`) means consumers can keep using `AgentId` interchangeably as a type or value, no name collision.

### 5. Scope discipline — PASS

`git diff aae3dbfd..570d8022 -- packages/server/src/app/app-host.ts packages/server/src/identity/services/agent-visibility.ts packages/server/src/network/services/presence.service.ts packages/server/src/task/services/message.service.ts` is **empty**. The 5 other `as AgentId`/`as UserId` sites at:

- `app-host.ts:620 → return conn.auth.agentId as AgentId`
- `agent-visibility.ts:47 → restrictTo as AgentId[]`
- `presence.service.ts:139 → agentId: agentId as AgentId`
- `message.service.ts:540, :624 → ) as AgentId[]`

are intentionally left in place per the dispatch. CHANGELOG notes them as a separate follow-up. Good scope hygiene.

### 6. No regression to round-1/round-2 items — PASS

Diff is 4 files: validator, validator-test, identity barrel, CHANGELOG. HMAC byte-exactness, shared semaphore, dispatcher lifecycle live in unrelated files (no diff). Body drain in the validator is preserved verbatim:

```
.pipe(
  Effect.tap((response) => response.text),
  Effect.flatMap(HttpClientResponse.filterStatusOk),
  ...
)
```

Comment explaining the `Effect.cached` reuse on 2xx is preserved.

---

### CI snapshot

- `ci` SUCCESS (3m34s)
- `test` pending
- `conformance` pending
- `Mintlify Deployment` skipped

No CI red. Pending checks should land before merge.

### Outcome

The fix correctly converts the two `as AgentId` / `as UserId` casts into a runtime brand decode that enforces `format: "uuid"`, flows failures through the existing fail-closed handler, and ships with three non-vacuous tests. Supporting barrel change is convention-aligned and safe. Scope is disciplined.

**STATUS: clean — approve and merge after pending CI clears.**

Issue #718, opened 2026-05-28
Model conversation subscription as a first-class server-side index (retire per-connection conversationIds Set)


chughtapan:
## Intent

Replace the denormalized per-connection `conversationIds: Set<ConversationId>` cache with a first-class server-side subscription/membership index.

## Motivation

`conversationIds` on each connection is a denormalized read-cache of conversation membership, used by `network-send.ts → connectionCanReceive` to gate message fan-out without a DB query per message. Authoritative store is the DB (`conversation_participants`); the connection set is a cached copy hydrated on connect (`connect.handlers.ts → hydrateConnectionState`), `.add`'d on subscribe (`connection.ts`), `.delete`'d on `conversation.service.ts → removeParticipant`.

**The smell:** membership is scattered as mutable copies across connections. Every membership change must loop over all of an agent's connections and mutate each cached `Set` (literally what `removeParticipant` does — `getByAgent` then clear).

**Why now:** the multi-connection-per-agent fix (#712 / PR #716, codex r5 P2) made this materially worse — an agent with N live connections (web + CLI + mobile) now carries N duplicate `conversationIds` sets that must all stay in sync. This is the same duplicated-mutable-state shape that produced the #716 presence bug.

## Proposed shape

A first-class subscription index owned by one service (candidate: `ConnectionManager` or a dedicated `ConversationSubscriptionService`):
- `conversationId → Set<ConnectionId>` (or `agentId → Set<ConversationId>`), single source of truth, maintained on connect / join / leave / disconnect.
- Fan-out queries the index once per message instead of iterating connections and checking per-connection sets.
- Removes the per-connection `conversationIds` field entirely → no cross-connection sync burden, no duplication under multi-connection.

Open design questions for the contract/architect cycle:
- Index keyed by conversation (fast fan-out) vs by agent (fast membership-change) vs both.
- Query-at-delivery against `ConversationService` vs maintained in-memory index (perf vs consistency tradeoff).
- Interaction with the presence projection's own connection tracking (avoid a second parallel index).

## Scope / sequencing

- **Orthogonal to #705/#717** (apps-as-principal). #717 carries `conversationIds` forward unchanged (just branded `Set<ConversationId>`, with the dead `mutedConversations` field deleted). Do NOT fold this redesign into #717.
- Should land AFTER #717 (D+C) merges, so it builds on the three-arm `Connection` union + multi-connection model.
- Run as its own `/safer:contract → architect → implement` cycle.

## Note on mutedConversations

[...]

Issue #685, opened 2026-05-23
Client-side DM dedup: implement proactive list+reuse; remove dead reactive branch (post-#677)


chughtapan:
## Context

#677 retired the server-side `DEFAULT_APP_ID` `TaskRequest` dedup (`maybeTaskCreateDedup`). The intended replacement — the "one DM per pair" UX — was to move to the client as *list existing DMs → filter by participant set → reuse-or-create*. That client-side dedup is **not implemented**, and the old reactive client code is now stale/dead. Surfaced while driving Spec D3 PR #669 (epic #602).

## Problem A — dead reactive branch + stale reference (drift)

`packages/client/src/cli/commands/start.ts → createTaskAtomic` still has:

```ts
if (result.conversation === null) {
  // Dedup hit (per `tasks.handlers.ts → maybeTaskCreateDedup`): ...
  return { kind: "dedup", task: result.task } as const;
}
```

- The cited `tasks.handlers.ts → maybeTaskCreateDedup` **no longer exists** (removed in #677).
- `start.ts` always passes a defined `initialConversation` (`{ name }` or `{ name, participants }`), and the D3 `task/request` handler (`packages/server/src/app/handlers/task-request.handler.ts → handleAccept`) only returns `conversation: null` when `initialConversation` is *undefined*. So for the `start` flow this branch is **unreachable dead code**.
- `findReusableConversation` / `pickReusableFromPage` / `DEDUP_LOOKUP_PAGE_SIZE` / `DEDUP_LOOKUP_MAX_PAGES` in `start.ts` are only reachable from this dead branch — so they're dead too.

## Problem B — proactive client dedup unimplemented

The real "one DM per pair" UX (list the caller's DMs, filter by `{caller} ∪ invitedAgentIds`, reuse an existing DM conversation instead of creating a duplicate) does not exist. Two server integration tests hold placeholders:

- `packages/server/src/__tests__/integration/task/task-stamp-and-dedup.test.ts:23` — `vit.todo("client-side DEFAULT_APP_ID dedup — list + match")`
- `packages/server/src/__tests__/integration/task/task-conversation-family.test.ts:159` — `vit.todo("client-side DEFAULT_APP dedup — list + match")`

with the note: *"Server dedup retired; the 'one DM per pair' UX moves to clients (list + filter + create-or-use). Re-add coverage in the SDK package."*

## Fix recipe (separate PR)

[...]

Issue #681, opened 2026-05-22
D3 follow-up: replace 11 vit.todo markers in integration/task/* with bodies


chughtapan:
## Context

PR #669 (D3 Spec cutover) lands with 11 `vit.todo` markers across 5 integration test files. Per user verdict 2026-05-22 on parent epic #602 (issuecomment iteration-8 thread), these are accepted as scoped-out for the D3 cutover to keep the cross-layer collapse atomic. This issue captures the deferred work so it does not get lost.

## Scope

Replace `vit.todo` markers with actual test bodies. Two recurring blockers, both inside `packages/server/src/__tests__/integration/task/`:

1. **AppsRegister fixture** — `conversations-archive.test.ts` (5 markers) + `group-chat.test.ts` (1 marker) cannot exercise the archive/group-chat paths without a test fixture that registers apps in the same shape D3 expects post-collapse.
2. **Client-side DEFAULT_APP_ID dedup** — `task-conversation-family.test.ts` + `task-stamp-and-dedup.test.ts` (1 each) need a client-side dedup oracle (list + match) covering the renamed DEFAULT_APP_ID introduced in PR #669 commit `5065111` (the AppHost-into-AppRegistry collapse).

The remaining 3 markers (`task-manager-routing.test.ts`) cover TM routing edge cases — read the file to identify exact scenarios.

## Inventory

| File | Count | Theme |
|---|---|---|
| `packages/server/src/__tests__/integration/task/conversations-archive.test.ts` | 5 | needs AppsRegister fixture |
| `packages/server/src/__tests__/integration/task/group-chat.test.ts` | 1 | needs AppsRegister fixture |
| `packages/server/src/__tests__/integration/task/task-conversation-family.test.ts` | 1 | client-side DEFAULT_APP dedup |
| `packages/server/src/__tests__/integration/task/task-manager-routing.test.ts` | 3 | TM routing edge cases |
| `packages/server/src/__tests__/integration/task/task-stamp-and-dedup.test.ts` | 1 | client-side DEFAULT_APP_ID dedup |

Enumerate by grep on PR #669 branch `impl/600-d3-cutover` post-merge:
```bash
grep -RnE 'vit\.todo\(' packages/server/src/__tests__/integration/task/
```

## Acceptance

- Every `vit.todo` from the inventory above is replaced with a passing test body OR explicitly converted to `it.skip` with a justification comment naming why the scenario is structurally unreachable at HEAD.
[...]

Issue #670, opened 2026-05-21
Perf: TaskCreate dedup is O(N) round-trips and double-runs contact-policy


chughtapan:
Two pre-existing perf issues surfaced during /simplify review of PR #669 (D3 cutover), now load-bearing post-cutover:

## 1. `findExistingTaskByParticipants` is N+1 (server-side)

**File:** `packages/server/src/task/services/task.service.ts → findExistingTaskByParticipants` (called from `tasks.handlers.ts → maybeTaskCreateDedup`).

For a creator with K prior tasks under `DEFAULT_APP_ID`, the helper issues:
- 1 query for `candidateTaskIds`
- per candidate (sequential, no `Effect.forEach({ concurrency })`):
  - 1 query for participant fan-out
  - 1 `selectAll` on `tasks`

= **2K + 1 round-trips per `task/create` invocation**.

Pre-D3 this path was less hot because conversation-level DM dedup (`findExistingDm`, deleted in a48a0d1) handled the `messages/send to: agent:<name>` shortcut. Now this is the single TaskCreate dedup site.

**Fix sketch:** collapse to one SQL query joining `tasks` + `task_participants` with `HAVING COUNT(*) = $size AND BOOL_AND(agent_id IN (...))` (or `GROUP BY task_id HAVING array_agg(agent_id ORDER BY agent_id) = $expectedArray::uuid[]`). PRIMARY KEY `(task_id, agent_id)` covers the inner filter.

## 2. Contact-policy gate runs twice on `task/create` + `initialConversation`

**File:** `packages/server/src/task/handlers/tasks.handlers.ts → taskCreateBody` + `mintInitialConversation`.

The handler runs:
1. `maybeRunContactPolicyForTaskCreate` → `obtainContactPolicyForCreate` → `loadAgentOwners` IN-scan + `assertCreatorContactsAll` (creator row read + per-target `checkContactEdge`)
2. `mintInitialConversation` → `obtainConversationCreateAuthorization` → re-runs `loadAgentOwners` + `assertContactPolicyForCreate` (plus group-capacity)

= **2× loadAgentOwners + 2× creator-row read + 2× per-edge policy probe** on the create-with-initial-conversation path.

**Fix sketch:** have `mintInitialConversation` provide a no-op / cached `ConversationCreateAuthorization` (the gate already ran upstream), or thread the cached `ownerByAgentId` map through. Requires the obtain helper to accept a "policy-already-ran" branch.

## Why not fixed in #669

[...]

PR #722, opened 2026-05-29
feat(#705 D+C): apps as first-class principals -- cutover

### LINKED ISSUE #478: Unify notification fan-out helper across MessageService + handlers
## Context

PR #475 (Phase 10 Broadcaster removal) folded fan-out onto `NetworkSendService.broadcast`, but the `notificationFrame → JSON.stringify → opaquePayload → broadcast` chain is now hand-rolled at five sites:

- `packages/server/src/task/handlers/conversations.handlers.ts:39-58` — local `broadcastToConversation` helper
- `packages/server/src/task/handlers/conversations.handlers.ts:87-95` — inline `ConversationsCreate` fan-out
- `packages/server/src/task/handlers/contacts.handlers.ts:40-54` — `fanOut` helper
- `packages/server/src/services/message.service.ts:283-294` — inline broadcast in `MessageService.send`
- `packages/server/src/services/presence-event-sink.ts:31-50` — predates Phase 10; still hand-rolls a per-connection loop with manual `Effect.runFork`

## Fix recipe

Add a `broadcastNotification<D extends NotificationDefinition<string, TSchema>>(agentIds, definition, params, opts?)` method on `NetworkSendService` (or a free helper next to `opaquePayload`). It owns the frame + JSON + brand + fan-out chain. Every call site collapses to one line.

Pair it with a `sendNotification(to, definition, params)` overload for the directed-send case at `message.service.ts:263-269` (the TM-routing send).

`presence-event-sink.ts` then migrates onto `NetworkSendService.broadcast` and the manual fork loop disappears.

## Why deferred from #475

PR #475 scope is "delete Broadcaster + migrate consumers." Helper unification is additive cleanup that touches an out-of-scope file (`presence-event-sink.ts`) and is best done as a follow-up so the Phase 10 review surface stays focused.

## Acceptance

- One helper signature; five sites collapse onto it.
- `presence-event-sink.ts` no longer hand-rolls connection iteration + `runFork` (delegates to `NetworkSendService.broadcast`).
- No new public exports — the helper is internal to `network-send.ts` (or a sibling file under `network/`).

### LINKED ISSUE #500: protocol: vestigial wire methods + notifications without production sender/handler
## Surfaced by

`/simplify` PASS 3 of STAMINA on PR #493 (Phase 12 protocol finalization).

## Three vestigial wire-registered defs

Each is registered in `identityRpcMethods` / `taskNotifications` (so the wire dispatcher would route it), but no production code emits or handles it.

### 1. `agents/invite` (`InviteAgent`)

- Defined: `packages/protocol/src/identity/methods.ts:149-156`
- Registered: `identityRpcMethods` (same file, line 281)
- Consumers: only `packages/protocol/src/testing/models/dispatch.ts` (test reference model, lines 26 + 203)
- Server: no handler. Wire request would resolve to `MethodNotFound`.

### 2. `invites/createAgent` (`InvitesCreateAgent`)

- Defined: `packages/protocol/src/identity/methods.ts:273-279`
- Registered: `identityRpcMethods` (line 292)
- Consumers: only `testing/models/dispatch.ts` (lines 57 + 258)
- Server: no handler.

### 3. `task/failed` (`TaskFailedNotificationDefinition`)

- Defined: `packages/protocol/src/task/methods.ts:677-680`
- Registered: `taskNotifications` (line 715)
- Re-exported from: `packages/protocol/src/testing/index.ts:30` (NOT from root `src/index.ts` nor `task/index.ts`)
- Consumers: only `packages/client/src/runtime/subscribers.test.ts` (test for subscriber dispatch shape)
- Server: no notification emitter; nothing fires `task/failed`.
- Client: no subscribe handler.

## Resolution options per def

For each:
- **(a) Wire it up**: implement server emitter / client handler against the existing schema (matches the spec intent)
- **(b) Delete it**: remove def + array entry; consumer test models adjust

## Why this slipped past knip (#496)

Knip flags symbols whose `export` site has zero references. These defs ARE referenced — by `identityRpcMethods` / `taskNotifications` registry arrays — so they look "used" structurally. The dead-end is at the wire/handler layer, which knip doesn't trace.

## Recommendation

Triage option (a) vs (b) per def with the owning service team. The longer they sit registered without a handler, the more confusing the wire surface becomes for consumers reading the registry.

### LINKED ISSUE #702: Track B: listGroupsEffect + service.ts consume TaskConversationList single-page (silently truncate >100)
Surfaced by review-senior during PR #700 (Track A pagination) round-2 review; **out of Track A scope** (Track A only paginated `agents/list`, `contacts/list`, `task/list`).

## The latent bug
Two production consumers read `TaskConversationList` (`task/conversation/list`) **single-page**, ignoring its `nextCursor`, so they silently truncate at the cap (100):
- `packages/openclaw-channel/src/openclaw-entry.ts → listGroupsEffect` — the directory's group enumeration.
- `packages/client/src/service.ts` (~the `TaskConversationList` consumer) — conversation-metadata path.

`task/conversation/list` is the legacy cursor surface (plain `Type.String()` cursor, cap 100), landed in #669/D1 — predates Track A, so it was correctly left untouched there.

## Why this belongs to Track B (#697)
Track B **retires `TaskConversationList`** in favor of the unified `TaskList` item (`taskParticipants` + caller `conversations`). When Track B migrates these two consumers off `TaskConversationList`, it MUST ensure the replacements **paginate to completeness** (mirror Track A's `drainPaginatedList` drain-all pattern) — i.e., do not carry the single-page truncation into the new surface. This is a checklist item for the retirement, not a standalone fix.

Parent epic: #693 · Track B: #697

### LINKED ISSUE #704: Retire in-process default-app loopback: default app as MoltZapTMClient over localhost socket + compile-time agent/app sendRpc split
Parent epic: #693 · **Blocks: #697 (Track B)** · Ships FIRST, before Track B.

## Why this is carved out of Track B
Surfaced designing Track B (#697) DM dedup. The default app (TM for `DEFAULT_APP_ID`) runs **in-process** via `makeLoopbackConnection` (`packages/server/src/app/loopback-connection.ts`, wired by `installDefaultApp` in `default-app.ts`) — a one-way server→app callback shortcut (`originator.call` invokes the handler directly; `handle`/`resolve` defect). Track B needs the default app to ALSO *call* server RPCs (for dedup), and v3–v5 tried to bend the loopback two-way (`setLoopbackInbound`, a `systemApp` ctx-marker, an agent-less dispatch carve-out) — over-engineered.

**Insight:** `MoltZapTMClient` (`packages/client/src/tm-client.ts`) already exists — connects over a socket, has `sendRpc` (outbound server RPCs) AND a `TMHandlers` table (receives TM callbacks). **Wire TMs already use it.** The default app is the *only* TM on the in-process loopback.

## Goal (behavior-preserving refactor, ship first)
1. **Default app becomes a real `MoltZapTMClient`** over a **real localhost socket** (transport option 2, user-chosen): connects after the server starts listening (boot ordering), registers as `DEFAULT_APP_ID`, with today's callback handlers UNCHANGED (`dispatch/authorize` → grant, `messages/authorize` → forward, `task/create` → accept). No dedup yet.
2. **Delete the in-process loopback dispatch special-case** (`makeLoopbackConnection`'s short-circuit + any two-way plans). The default app uses the identical path wire TMs do.
3. **Compile-time agent/app `sendRpc` separation:** type each client's `sendRpc` by descriptor layer so a `MoltZapAgentClient` **cannot** call an app-layer RPC (no runtime `ForbiddenError`). This is the typed foundation Track B's `app/tasks/list` will rely on.

## Out of scope (stays in Track B #697)
`app/tasks/list`, DM dedup, the unified `TaskList` item / `taskParticipants`, `TaskConversationList` retirement, the per-pair serializer + concurrency proof.

## Accepted tradeoff (user decision)
The co-located default TM now pays a localhost socket round-trip per callback (vs zero-overhead in-process) — accepted for uniformity (default app == wire TM, one code path).

## Verify during design
- Confirm a registered app connection (auth model for apps — likely `auth: null` but registered) can `sendRpc` server RPCs today, so the default-app-as-TMClient uses the identical authorized path.
- Boot ordering / self-connection (server's own process opens a client to its own listener).

### LINKED ISSUE #717: impl(#705): D+C co-merge — apps as first-class principal + loopback retirement (v36 plan)
## Scope

Implement the D+C co-merged plan from #705 v36 (final consolidated spec). Per v18 USER directive, C's loopback retirement (#704) is absorbed into D's PR — single atomic deliverable.

## Plan reference

#705 v36 (final consolidated). Comments 4566573737–4566577239 (18 parts, 550 KB).

Authoritative source-of-truth. Architecture has been settled since v22 (10+ rounds of 0 P1 from plan-eng).

## Branch base — STACKED ON #716

**Branch from `impl/706-presence-projection`** (current PR #716 head), NOT from `main`. When #716 merges to main, rebase this branch onto main; conflict-free because #716 lands the lease-registry vestigial sweep + presence projection that D+C work depends on.

PR opens against `impl/706-presence-projection` as base. When that's rebased post-#716-merge, target switches to main and merges cleanly.

## Implementation phases (per v36 §9.2 file-edit table)

This is a LARGE PR (~10-12d human / ~22-26h CC per v36 §15 Gate Effort). Headline work:

1. **Three-arm `Connection` discriminated union** — `UnauthenticatedConnection | AgentConnection | AppConnection`. Module-private constructors with nominal `__brand: never` seal. Delete `MoltZapConnection.auth: AuthenticatedContext | null`.
2. **Handler migration** — 32 `ctx.agentId` reads across 7 server handlers + 4 protocol capability-providers (`messages.ts:208/250`, `tasks.ts:417/494`) rewrite to `conn.auth.agentId`.
3. **Default-app rewrite (C absorbed)** — DELETE `loopback-connection.ts` (144 lines). REWRITE `default-app.ts`: `installDefaultApp` → `startDefaultApp`; constructs real `MoltZapAppClient` over loopback; **passes `reconnectPolicy: "none"`**.
4. **AppRegistry surface** — NEW `lookup(appId)` with `bootCompleteAt: number | null` discriminator + `markBootComplete()`.
5. **Boot lifecycle** — `Scope.addFinalizer(appScope, dispatchRuntime.dispose)`. serverSvc wrapped in `Scope.extend(appScope)`. Step 5a/5c `Effect.mapError` envelopes with `ServerBootFailedError({phase})`. **`logAndSwallowCause` helper** at 7 cleanup positions.
6. **Default-app no-reconnect** — `AppClientOptions.reconnectPolicy?: "exponential-backoff" | "none"`.
7. **HelloOk empty result + HTTP route flag + AlreadyConnected typed error + safeEqual + wire codes** -32006/-32011/-32012.
8. **HTTP `POST /api/v1/apps/register`** — server-issued `appId` via `gen_random_uuid()`.

## Substrate-citation correction (plan-eng r35 finding)

v36 §9.2 cites "AppHost.mintDispatchLease" but the actual symbol is `registry.mint({...})` inside `enqueueDispatchRequestEffect` at `app-host.ts:401`. Verify against HEAD before implementing the dispatch-lease path.

## Acceptance criteria

Per v36 §15 acceptance bullets. Notable items:
- chughtapan/moltzap#715 exists (gate-split follow-up)
- Three-arm union mechanically enforced: AUTH gate is `conn._tag === "UnauthenticatedConnection"`
- Default-app `MoltZapAppClient` constructed with reconnect disabled (code-level grep verifiable)
- Binding-asymmetry invariant: `expireLeaseOnDisconnect` keyed on `recipientConnectionId` only
- §15 audit script #1-#15 all PASS

## Out of scope

- Codex r29-r34 peripheral lifecycle items not in v36 — pick up at /simplify + /review pre-PR gates
- #710 / #711 / #712 / #716 (separate workstream, already in flight)
- B #697 (depends on D, not in scope here)

## Discipline

- /safer:simplify + /safer:review pre-PR (mandatory per `feedback_simplify_review_mandatory`)
- /document-release pre-merge (mandatory per `feedback_impl_runs_document_release`)
- NO deferred cutovers (memory `feedback_no_defer_finish_in_impl`)
- NO `--no-verify` (`feedback_no_skip_hooks`)
- NO raw SQL (`feedback_no_raw_sql`)

## Process

1. Create branch `impl/705-d-plus-c-cutover` from `impl/706-presence-projection`
2. Implement against v36 spec
3. Open draft PR against `impl/706-presence-projection` as base
4. When #716 merges, rebase to main; PR base auto-updates
5. Six pre-push gates exit 0 each push
6. Land — user-gated

🤖 Filed by orchestrator after #705 advanced to plan-approved.

### LINKED ISSUE #720: Agent-only RPCs lack a runtime principal-kind gate against AppConnection callers
## Context

Surfaced by codex during pre-landing review of the #705 D+C cutover PR (branch `impl/705-d-plus-c-cutover`).

The cutover introduces `AppConnection` as a first-class WS principal (appKey Connect). A wire app client can therefore send ANY frame, including agent-only RPCs such as `messages/send` and `messages/list`.

## The gap

`packages/server/src/transport/context.ts → defineMethod`, the `requiresActive` gate:

```ts
const auth =
  ctx.connection._tag === "AgentConnection"
    ? ctx.connection.auth
    : ({} as AuthenticatedContext);
if (
  requiresActive &&
  ctx.connection._tag === "AgentConnection" &&   // <-- app arm skips the gate
  auth.agentStatus !== "active"
) { ... ForbiddenError ... }
```

For an `AppConnection`:
- the `requiresActive` gate is skipped (it only fires on the agent arm), and
- the handler runs with `ctx.agentId === undefined` (the `{} as AuthenticatedContext` placeholder).

So `messages/send` reaches `messageService.send({ senderAgentId: undefined, ... })`.

## Severity: fail-closed today, robustness gap

This is NOT a live authorization bypass: `messages/send` is gated by the `MessageSendPermission` composite capability, which requires an admitted participant. `senderAgentId: undefined` matches no participant, so the send fails closed (ForbiddenError/ParticipantNotAdmitted). Verified the downstream capability gate fails closed.

The risk is structural: the safety relies on every agent-only RPC having a downstream participant/identity gate. Any current or future agent-only RPC without one would run with an undefined agent identity instead of cleanly rejecting the app principal. The `requiresActive` gate should reject the wrong principal kind at the boundary, not delegate fail-closed behavior to each handler.

## Why not a mechanical review fix

The binding layer does NOT correspond to the calling principal. Example: `task/request` (`packages/server/src/app/handlers/task-request.handler.ts`) binds via `defineAppMethod` (it needs `AppHostTag`) AND sets `requiresActive: true`, but it is AGENT-called and legitimately reads `ctx.agentId` as `initiatorAgentId`. So neither the bind layer nor the `requiresActive` flag currently encodes "which principal kind may call this RPC."

A correct fix needs a first-class **principal-kind dimension** on the method binding (agent-callable vs app-callable vs either), wired through the layer wrappers and enforced in `defineMethod`. That is new authority modeling — architect-tier, aligned with #705/#704.

## Fix recipe

1. Add a `callablePrincipal: "agent" | "app" | "any"` (or similar) to the method binding, defaulting to `"agent"` for the agent-only RPCs and `"app"`/`"any"` for app-arm RPCs.
2. In `defineMethod`'s slot handler, reject a mismatched principal with `ForbiddenError` BEFORE invoking the body — so `messages/send` from an `AppConnection` returns a clean 403 and never runs with an undefined `agentId`.
3. Audit the `requiresActive: true` + `defineAppMethod` set (`task/request`, `dispatch/request`, `dispatches/get`, the task-admin RPCs) and tag each with its true calling principal.

Citation: `packages/server/src/transport/context.ts → defineMethod` (the `requiresActive` block); `packages/server/src/transport/define-layered-method.ts` (layer wrappers that only gate the R-channel type).

### LINKED ISSUE #721: TMClientOptions requires agentKey even for app-principal clients
## Context

Surfaced by codex during pre-landing review of the #705 D+C cutover PR (branch `impl/705-d-plus-c-cutover`). Codex severity P2 (non-blocking ergonomics).

## The gap

`packages/client/src/tm-client.ts → TMClientOptions`:

```ts
export interface TMClientOptions {
  serverUrl: string;
  agentKey: string;   // required
  appKey?: string;    // optional, app-principal credential
  ...
}
```

`awaitConnectAuth` selects the credential: when `appKey` is set it sends the app arm and NEVER reads `agentKey`. But because `agentKey` is a required field, a wire APP client constructed from only the `{appId, appKey}` returned by `/api/v1/apps/register` must pass an unused dummy `agentKey: ""`.

## Why not fixed inline in the cutover review

A minimal "make `agentKey` optional + throw if neither/both" patch runs into the repo's `agent-code-guard/no-raw-throw-new-error` rule (raw `throw new Error` is banned in favor of tagged errors) and `no-nested-ternary`. Doing it cleanly means either a `Data.TaggedError` for the invalid-options case or modeling the credential as a discriminated union — codex's own suggestion. That is a typed-API shape change worth doing deliberately, not as a drive-by in the cutover.

## Fix recipe (preferred: discriminated union)

```ts
type TMClientCredential =
  | { readonly agentKey: string }
  | { readonly appKey: string };
export type TMClientOptions = TMClientCredential & {
  serverUrl: string;
  handlers: TMHandlers;
  // ...rest
};
```

Then `awaitConnectAuth` narrows on `"appKey" in this.options` (the union makes both arms type-safe with no assertion). All current call sites pass exactly one credential, so they migrate cleanly.

Citation: `packages/client/src/tm-client.ts → TMClientOptions` (the `agentKey: string` field) and `awaitConnectAuth` (the credential-selection ternary).

### LINKED ISSUE #723: Migrate wire validation TypeBox+AJV → Effect Schema (#705 Half 2)
# Migrate wire validation TypeBox+AJV → Effect Schema (#705 Half 2)

**Half 2** of the dispatcher re-type. Spike-proven (#720, the HALF-2 EFFECT-SCHEMA SPIKE comment): swapping the wire-validation engine from TypeBox+AJV to `effect/Schema` sidesteps **TS#30581** (a discriminated `WireFrame` `Schema.Union` + `Match.exhaustive` narrows `frame.params` per method **cast-free**, where the status-quo `Record<method, handler>` lookup re-hits the wall) **and** unifies the repo's four validation stacks onto one engine.

## Composition
- **Bases on** PR #722 (`impl/705-d-plus-c-cutover`, #705 D+C cutover).
- **Builds on** Half 1 (the cast-free closure-existential `ErasedSlot` + `DispatchContext` dissolution — architect plan on #720, in flight). Half 2 assumes Half 1 has landed and **subsumes** specific Half-1 constructs (the per-method AJV decode inside `ErasedSlot.invoke` and the runtime `table[method]` lookup) while **reusing** the rest (the cast-free slot existential, the capability discharge, the `#720` principal-kind gate).

## Scope (clean single-PR cutover — NO parallel-validator period)
Repo doctrine rejects coexistence shims. One PR swaps the engine end-to-end:
- P1 `schema-primitives.ts` brand/primitive factories + the 3 `FormatRegistry` checkers → `Schema`.
- P2 core wire (`method.ts`, `wire.ts`, both `Ajv` instances deleted, `ParseError` → wire codes).
- P3 dispatcher discriminated-union cutover (derive all-methods `WireFrame` `Schema.Union` from the registry; `Match.exhaustive` dispatch; delete `eraseHandlerTable`/`eraseProviderTable` + `table[method]`).
- P4 introspection consumers (`scripts/docs/schema.ts`, `testing/arbitraries/from-typebox.ts`, conformance re-coupling, the `Static<>` → `Schema.Schema.Type<>` canary sweep).
- P5 ~50 `defineRpc`/`defineNotification` call-site ports across the 8 production schema files.
- P6 per-inbound-frame micro-benchmark as a **merge gate** (the spike flagged `Schema.decode` per-frame interpretation vs AJV compiled validators as UNPROVEN).

## Public-surface blast radius
New `RpcDefinition` shape across protocol + server + tooling. Per the spike: **N=2 plan review** before implementation.

## References
- Spike verdict + evidence: #720 (HALF-2 EFFECT-SCHEMA SPIKE comment)
- Parent: #705 (apps-as-first-class-principals; D+C cutover = PR #722)
- Half 1 amendment: #720 (HALF-1 cast-free amendment comment)

Detailed phased architect plan posted as a comment below.

### LINKED ISSUE #725: Evaluate adopting @effect/rpc vs the custom dispatcher (#705/#723 upstream)
# Verdict: Evaluate adopting @effect/rpc vs the custom dispatcher

**Decision: PARTIAL.** Adopt `@effect/rpc` for the dominant client→server catalog (serialization + handler-Layer + typed-client + capability/principal middleware); KEEP a thin bespoke layer for the symmetric server→client callbacks and id-less notifications. **Do NOT** treat this as a free byproduct of #723 — it is its own architect-tier effort. **#723 (Effect-Schema migration) remains the immediate path and is NOT superseded.**

This synthesizes three independent probes (surface-mapping A, load-bearing-requirements B, runtime-POC C). All three converge on PARTIAL; all load-bearing claims below were re-verified against HEAD.

---

## What @effect/rpc would subsume

| moltzap layer | Subsumed? | @effect/rpc replacement |
|---|---|---|
| **Half 2 — Effect-Schema migration (#723)** | YES, for client→server direction | `RpcGroup.make(Rpc.make(tag,{payload,success,error,stream}))` IS effect/Schema request/success/error definition. This is literally the #723 migration for the dominant catalog. |
| **The AJV stack** (TypeBox + AJV validation) | YES, for client→server | `RpcSerialization` (json/ndjson/msgPack/jsonRpc) + Schema-derived encode/decode replaces TypeBox+AJV at that boundary. |
| **The client** (`RpcClient.make`, typed send/dispatch) | YES | `RpcClient.make` gives the typed client; clientId-keyed fan-out is built in. Proven by existing `LocalDaemonRpcs` integration (`packages/client/src/local-daemon-rpc.ts`, `packages/client/src/runtime/local-socket-server.ts`). |
| **Capability auto-provisioning** (`dispatch.ts → applyCapabilityProvisioning`) | YES (POC-proven) | `RpcMiddleware.Tag(name,{provides})` + `Rpc.middleware(M)`. Internal `RpcServer.applyMiddleware` does `Effect.provideServiceEffect(handler, tag.provides, middleware(options))` — structurally identical to moltzap's reverse-fold `provideOne`. POC1 proved params-derived provisioning, short-circuit on provider failure, and a compiler-native lockstep gate (unprovided tag leaks into RIn ≠ never) that replaces the hand-written `typed-dispatcher.types-check.ts` canary. |
| **Principal-kind gate (#720) + 3-arm auth context** | YES (POC-proven) | An `Authenticate` middleware that `provides` a tagged-union `CurrentPrincipal`. POC3 matrix: app→OK, agent-active→OK, agent-pending→ForbiddenError, unauthenticated→ForbiddenError — mirrors server `defineMethod`. **Correction to the original framing:** this gate does NOT live in the dispatcher — it is a runtime check inside handlers (`tasks.handlers.ts → assertCallerAppOwnsTask`, `connection._tag !== "AppConnection"`). The `agentCallable`/`appCallable` split (`task/methods.ts`) is a compile-time OUTBOUND catalog partition, also not dispatcher-coupled. So less is dispatcher-coupled than #720's framing assumes. |
| **Half 1 — ErasedSlot / HandlerSlot table** (`handlers.ts`) | PARTIALLY | `group.toLayer/toHandlers` replaces the static `HandlerSlot`/`HandlerTable` wiring for the client→server direction. The callback direction would still need a slot-like table. |
| **The wire envelope** (`wire.ts` JSON-RPC 2.0 + -32xxx codes + Connect version negotiation) | **NO — incompatible** | See gaps. `@effect/rpc`'s `layerJsonRpc` is JSON-RPC-FLAVORED, not wire-compatible. |

---

### 1. Server→client callbacks — CRITICAL BLOCKER (intrinsic protocol semantics)
moltzap runs a **symmetric** originator+dispatcher on BOTH ends of one WebSocket. Verified at HEAD:
- `protocol/src/transport/connection.ts → ServerConnection extends InboundDispatch + OutboundCall<AnyTaskCallbackRpcDefinition>` and `TaskMasterConnection extends InboundDispatch + OutboundCall<AnyTaskMasterRpcDefinition>` — each connection value carries BOTH directions.
- `protocol/src/transport/originator.ts → makeOriginator/callWithRefs` with a `Deferred`-keyed pending `HashMap` — the server issues a typed request INTO the client (`app-host.ts → dispatchAuthorizeHook` / `callAppRpc → sendRpcToClient`) and awaits a typed reply (`DispatchAuthorize`, `MessagesAuthorize`, `TaskCreate`).

`@effect/rpc` is strictly directional. Verified in dist dts: `FromClient = Request | Ack | Interrupt | Eof`; `FromServer = ResponseChunk | ResponseExit | ResponseDefect | ClientEnd` — **there is no server-originated Request arm.** `Protocol.send` only emits a response correlated to a prior client Request.

POC2 proved the ONLY way to express this is an **architecture inversion**: run a SECOND `RpcGroup` in reverse (network-server holds `RpcClient<CallbackGroup>`, the connected client runs `RpcServer<CallbackGroup>`), bridged onto one transport via `makeNoSerialization` + a custom `Protocol` frame-demultiplexer. This is **net-new transport plumbing, not a deletion** — it re-invents the bidirectional layer `@effect/rpc` deliberately omits.

### 2. The wire envelope — incompatible dialect (intrinsic protocol semantics)
moltzap's wire is plain JSON-RPC 2.0 with strict TypeBox `additionalProperties:false` frames, **string** ids, a registered error-code model (`wire-errors.ts`, codes -32000..-32024 carrying `{code,message,data?}`), and Connect-time version negotiation (`ProtocolMismatchError -32006`). `@effect/rpc`'s `layerJsonRpc` is a DIFFERENT dialect: it injects `@effect/rpc/`-prefixed control pseudo-methods (Ack/Interrupt/Ping/Eof), derives **numeric** ids (`id: Number(response.id)` — moltzap ids are strings), adds non-frame fields (`chunk:true`, `headers`, `traceId`), and wraps errors as `{_tag:"Cause"|"Defect", code, message, data:<serialized Cause>}` — which moltzap's `RpcErrorSchema` (`additionalProperties:false`) would REJECT. **`@moltzap/client` is consumed by moltzap-arena, so a wire-dialect change is a real external cutover**, not a no-external-consumers freebie.

### 3. Id-less notifications fan-out — semantic rewrite, not lift-and-shift
moltzap's `NotificationFrame` (no id, no response — `defineNotification`, `makeNotify`; ≥9 sites across task/identity/network/app; client `subscribers.ts` Stream fan-out). `@effect/rpc` models server push as streaming Rpc results (`Chunk`), not standalone id-less notification frames. Portable only by reshaping every notification into a long-lived stream subscription — a semantic rewrite.

### Not gaps (transport-agnostic, verified): the principal gate (#720) and lease/dispatch-admission (`LeaseRegistry`, `app-host.ts`) live in handlers/services, reachable via Context/middleware unchanged.

---

## Does this supersede #723? — NO

#723 replaces TypeBox+AJV with effect/Schema **at the same wire boundary**. An `@effect/rpc` move changes the wire boundary itself (new dialect) AND adds bidirectional-Protocol + notification plumbing + an arena cutover. They are orthogonal:
- #723 is a low-risk, in-place validation-stack swap on the existing symmetric wire.
- `@effect/rpc` adoption is an architect-tier transport replacement with a non-interoperable wire and net-new bidirectional plumbing.

**Do #723 as planned.** Doing #723 with effect/Schema also de-risks a later `@effect/rpc` move (the request/success/error Schemas become directly reusable as `Rpc.make` payloads), but #723 must not be paused or redirected.

---

## Migration size (if the PARTIAL adoption is later greenlit)

- **Client→server catalog → @effect/rpc** (serialization + handler Layer + typed client + capability/principal middleware): ~20-30 CC-hours. Blast radius: `protocol/transport/{method,handlers,wire,capabilities}.ts`, server `define-layered-method.ts`/`context.ts`, client `agent-client.ts`. The capability + principal machinery is the bulk; POC-proven to map.
- **Bidirectional callback layer** (the inversion or a retained thin originator): ~30-50 CC-hours and the GATING RISK. Requires a custom `RpcServer.Protocol`/`RpcClient.Protocol` pair demuxing two logical RPC endpoints on one socket. A spike to prove dual-endpoint-on-one-socket is mandatory before committing.
- **Wire-dialect cutover + arena** (`@moltzap/protocol`/`client` republish, arena upgrade): the external blast radius; sequence behind a green spike.
- **Total rough order:** 60-90 CC-hours, architect-tier, multi-PR. Not a #723-sized swap.

---

## Reroute Half 2? — NO. Roadmap impact

- **Half 1 (ErasedSlot):** proceed; the slot abstraction is replaced for the c→s direction only IF @effect/rpc is later adopted — independent of the #723 timeline.
- **Half 2 / #723:** **stays the path, unchanged.** Use effect/Schema so the request/success/error Schemas are reusable as future `Rpc.make` payloads.
- **#722 (apps as first-class principals — cutover) / #720:** unaffected by this decision. The principal gate is a handler/middleware concern that ports cleanly under either dispatcher; land #722/#720 on the custom dispatcher now.

---

## Recommendation

1. **Ship #723 on the custom dispatcher with effect/Schema.** Do not pause or redirect. (Subsumption of #723 by @effect/rpc is partial and not free.)
2. **Land #722/#720 on the custom dispatcher.** Their requirements port cleanly under either model; no reason to block on a transport decision.
3. **File the @effect/rpc adoption as a separate architect-tier track** (this issue), gated on a **mandatory spike** proving the dual-endpoint-on-one-socket bidirectional `Protocol`. That spike is the go/no-go for full adoption.
4. **The decisive constraint** is the symmetric dual-role-per-socket topology (`ServerConnection`/`TaskMasterConnection` each `extends InboundDispatch + OutboundCall`). `@effect/rpc` was not built for it; the callback path forces an inversion. Until the spike proves that pattern, KEEP the bespoke bidirectional layer.

Links: #723 (Half 2 Effect-Schema), #720 (principal-kind gate), #705 (apps as first-class principal), #722 (D+C cutover).

### LINKED ISSUE #734: Connect/hello handshake cleanup (post-#725 cutover)
Part of #732 (apps-as-first-class-principals consolidation sprint). **Post-cutover** — sequence after #725 lands (the @effect/rpc cutover rewrites the connect/hello path).

## Intent

Clean up the connect / hello handshake surface once the effect/rpc cutover settles it. These are debt items surfaced during the consolidation but deferred so they ride the cutover's final shape rather than fighting it mid-flight.

## Items

- [ ] `HelloOk` payload — collapse to an empty/minimal struct if the cutover removes its remaining fields (verify against the post-#725 handshake).
- [ ] `sessionToken` — knip flags it; confirm dead post-cutover and delete (+ all consumers).
- [ ] One-credential path — fold any duplicate credential-carrying fields into a single source once the principal/middleware Context lands from #725.

## Acceptance

- [ ] Connect/hello surface has no knip-unused exports.
- [ ] No dual credential paths; one source of truth for the connecting principal.
- [ ] Full CI green on `impl/705-d-plus-c-cutover`.

Exact item list to be confirmed against the post-cutover handshake before dispatch — do not start until #725 is merged into the consolidation branch.

### LINKED ISSUE #735: fix(#705): manifest hook policies must be required — close the fail-open dispatch_authorize hole
## Summary

The receive-side dispatch authorization gate is **fail-open on hook omission**, and the omission is reachable for **any** wire-registered app — not just the trusted boot default app. This is the "manifest-default fast-path" (`D #705 CP8`) on PR #722 (unmerged). Fix it **first-class before #722 merges**: make every manifest hook policy a **required discriminated union**, so "no policy" becomes unrepresentable — a **compile error** for in-code manifest constructors (the default app, fixtures, first-party SDK) and a **decode rejection** at the wire boundary — rather than a silently-synthesized `grant`.

## The hole (grep-verified at `eaed650c`)

`AppManifestSchema` (`@moltzap/protocol app/methods.ts`) makes `hooks` *and* each of `dispatch_authorize` / `message_authorize` / `task_create` independently `Schema.optional`. `AppHost`'s three hook runners each branch on `entry.manifest.hooks?.<hook> === undefined` and synthesize the documented default **in-process, no app round-trip**:

- `dispatchAuthorizeHook` → `{ decision: "grant" }`
- `runMessageAuthorize` → `defaultMessageAuthorize` → `Forward { participants ∖ sender }`
- `runTaskCreate` → `{ decision: "accept" }`

`validateAppManifest` is shape-identical for boot and wire apps; `AppRegistry.register` has "no notion of 'boot' vs 'wire'." Registration is `POST /api/v1/apps/register`, gated only by an **optional** `registrationSecret` that is a **no-op when unset**. So an arbitrary wire app that omits `dispatch_authorize` (or the whole `hooks` object, or sends `hooks: {}`) **silently receives grant-all**.

### Why this is a posture contradiction
In the *same* `dispatchAuthorizeHook` method:
- registered app whose WS drops → `deny app_unavailable` (**fail-closed**)
- registered app that omitted the key → `grant` (**fail-open**)

Adjacent branches, separated only by whether a key is present. A **typo'd or dropped `dispatch_authorize` key downgrades a gated app to ungated with zero error, zero log, zero compile failure** — exactly the hazard the "hooks fail-closed" invariant exists to prevent. The policy is a *hole the server fills*, duplicated as prose in JSDoc (prose-as-contract that drifts).

## Why it was built this way (context, not justification)
Two legitimate goals (commit `9082dfd0`): (1) honor the `AppManifestSchema` JSDoc defaults the code never wired; (2) prerequisite for deleting the in-process loopback — a hookless default app needs no callback handler, so `loopback-connection.ts` can die. The default app's endpoint is genuinely **inert** (`makeDefaultAppEndpoint` → `Effect.die`), so synthesizing *its* verdict server-side is a necessity. The defect is that the mechanism is **general** (serves any omitting app) and **optional** (omission is valid), not that the default app is served in-process.

## First-class fix (no optional paths)

Make policy a **required discriminated union** on every hook slot — no `Schema.optional`:

```
dispatch_authorize: { kind: "grant" } | { kind: "deny"; reason: string } | { kind: "hook"; timeoutMs: number }
message_authorize:  { kind: "forwardAllExceptSender" } | { kind: "deny"; reason: string } | { kind: "hook"; timeoutMs: number }
task_create:        { kind: "accept" } | { kind: "reject"; reason: string } | { kind: "hook"; timeoutMs: number }
```

- **Omitting a policy is a compile error for authored manifests, a decode rejection at the wire.** A required (non-optional) field is enforced at BOTH ends from one definition: in-code manifest constructors (the default app, fixtures, first-party SDK) get `Schema.Type<typeof AppManifestSchema>` with the field **non-optional** → omission *fails to compile* (Principle 1, compile-time-first); wire manifests are rejected at decode (Principle 2, boundary backstop). "Forgot to declare a policy" is never a silent runtime grant. **Adding a new hook adds a new required field**, so every in-code constructor fails to compile until it declares the new policy — exhaustiveness at the *construction* site, not just the evaluation site (Principle 4 on both ends). The fail-open path stops being *representable*; the hole closes by **deletion, not documentation**.
- **One uniform evaluation for every app**: exhaustive `switch (policy.kind)` with **no `default:`** (a future kind → compile error, Principle 4). `kind:"grant"`/`"deny"`/`"accept"`/`"reject"`/`"forwardAllExceptSender"` are static verdicts resolved in-process (same zero-wire cost the fast-path has today); `kind:"hook"` is the **only** arm that touches the wire and the **only** arm under the fail-closed timeout envelope. No special boot-default lane.
- **The boot default app explicitly declares** `{kind:"grant"}` / `{kind:"forwardAllExceptSender"}` / `{kind:"accept"}` in `DEFAULT_APP_MANIFEST`. Its open posture becomes a grep-able, audited fact. It still needs no real callback handler → the loopback deletion goal is preserved.
- Unifies the two axes: "no policy" is gone (decode rejects); the only remaining absence-of-answer is **runtime hook failure → deterministic deny**. `deny` is never a fallback synthesized far from the policy site — it is an explicit `kind:"deny"` or the envelope's collapse of a failed `kind:"hook"`.

## Scope
- `@moltzap/protocol app/methods.ts` — `AppManifestSchema` + `HookEntrySchema` (fold `timeout_ms` into the `hook` arm); JSDoc rewritten to state the required-policy contract present-tense.
- `packages/server/src/app/app-host.ts` — collapse the three-arm `?.<hook> === undefined` branches into one exhaustive `switch (policy.kind)`; `defaultMessageAuthorize` becomes the `forwardAllExceptSender` arm. Drop the synthesized-default fast-path special-casing.
- `packages/server/src/app/default-app.ts` — `DEFAULT_APP_MANIFEST` declares all three policies explicitly.
- `validateAppManifest` consumers, conformance, unit tests, and any manifest constructors (check arena's app-sdk usage).
- Strip the `D #705 CP8` / spec-label provenance comments (cold-reader rule) in the touched blocks.
- No external consumers of the protocol → no back-compat shim; update all call sites in one PR.

## Process
Public-surface schema change on a security gate → architect plan → N=2 cross-model review → implement-staff. Lands on #722 **before merge** (user decision, 2026-06-01). Behavior-preserving for the default app (same grant / forward-all-except-sender / accept verdicts) and for wire apps that declare hooks. Parent epic: #732.

### LINKED ISSUE #736: Dead config knob: `log_level` is validated + documented but read nowhere
## What

The `log_level` key in the standalone YAML config (`server-core` `config.ts → MoltZapConfigShape`) is validated (accepts `debug|info|warn|error`, rejects typos) and documented to operators in `moltzap.example.yaml` (`log_level: info`), but **nothing reads it**. Setting it has zero effect on the server's log level.

## Evidence

- `config.ts` validates `log_level` but the decoded `YamlConfig` value is never projected into `StandaloneBootPlan`, and no logger setup reads it. Grep for `log_level` in `packages/server/src` returns only the schema + the test that asserts typo-rejection.
- The server's actual minimum log level comes from a different mechanism: `packages/client/src/cli/runtime.ts → minLogLevel`, mapped from the `MOLTZAP_LOG_LEVEL` env var (`Config.string("MOLTZAP_LOG_LEVEL")`).
- Writers of `log_level` into real config files:
  - `packages/nanoclaw-channel/vitest.integration.globalSetup.ts` (`log_level: warn`)
  - `packages/claude-code-channel/vitest.integration.globalSetup.ts` (`log_level: warn`)
  - `moltzap.example.yaml` (`log_level: info`)

So an operator who sets `log_level: debug` in their `moltzap.yaml` gets silent no-op behavior; the only way to change logging is `MOLTZAP_LOG_LEVEL`.

## Why it matters

The config lies to operators. The example file presents `log_level` as a working knob. The validator enforces its enum (reinforcing "this is real"), yet it does nothing.

## Fix options (decide, then route)

1. **Wire it up** — project `log_level` into the boot plan and feed it to the Effect logger's minimum level in standalone, so the documented knob works. (feature; `architect` → `implement-*`)
2. **Remove it** — drop `log_level` from the schema, from `moltzap.example.yaml`, from the two integration globalSetups, and the typo-rejection test; document `MOLTZAP_LOG_LEVEL` as the sole control. (cleanup; touches two channel packages + operator docs)

Surfaced during the typebox→Effect-Schema config-validator collapse (`cleanup/config-typebox`). The collapse preserved current behavior (still validates `log_level`); this issue tracks the underlying dead-knob decision, which is out of that PR's scope.

### LINKED ISSUE #739: fix(#705): unit test + conformance gates red — notification fan-out via s2c originator breaks unit/conformance harnesses
## Summary

On `impl/705-d-plus-c-cutover` (#722), the unit `test` gate (`@moltzap/server-core:test`: **36 failed / 183 passed**) and the `test:conformance` gate (**27 failed / 18 passed** server-side) are red. These failures are **pre-existing on the branch** — identical counts with and without the integration-greening commit `e7ad6ed7` (verified by stash-and-compare). The integration suite (`test:integration`) is now fully green (142/0).

## Root cause

The native-mux rearch routes server-originated notifications through the per-connection s2c reverse `RpcClient` (`originator.notify`) as **void-result RPCs that await the client's ack**, rather than a direct `socket.write`. The unit + conformance harnesses were built against the old direct-write path:

- **`presence.service.test.ts` (8 failed):** `makeHarness` seeds an unauthenticated connection with a recording `socket.write` but no `originator`. The presence fan-out (`broadcastNotificationToAgents`) now goes via the originator, so `h.sent` stays empty (`expected [] to deeply equal [{agentId, status:'online'}]`).
- **`app-host.test.ts` (1 failed):** same class — the mock connection lacks the s2c notify path.
- **dispatch-flow unit tests + `server-executable.proofs.test.ts` / `dispatch-admission-bad-server-executable.proofs.test.ts` (notification/broadcast + dispatch proofs):** 20s timeouts. The integration suite hit the SAME hang; the fix there (driver acks the s2c notification-RPC — commit e7ad6ed7) lives in the integration helpers/`test-client.ts` but the conformance proof harness + the unit mocks still don't ack, so the server's `notify` parks.

The integration commit fixed the driver-side ack (`test-client.ts → handleNotificationRequestFrame`) and the safe-integer request-id, which is why integration is green; the unit mocks and (some) conformance proof paths need the equivalent treatment.

## Scope of fix

1. **`presence.service.test.ts` / `app-host.test.ts` mocks:** record frames off the s2c originator (or have the mock originator's `notify` settle + capture), not just `socket.write`.
2. **Conformance proof harness:** ensure the proof TestClient acks server-pushed notification-RPCs (the `test-client.ts` fix should cover the shared driver; verify the `__divergence_proofs__` executable harness uses it and isn't on a divergent mock).
3. Re-run `nx run-many -t test test:conformance` to confirm green.

## Also pre-existing (separate, smaller)

- `native-mux.test.ts` — 2 encoder property tests fail (`JSON parser produced a non-string frame on channel s2c`) in `makeEnvelopeEncoder`; new test file, never green on the branch.
- `test-client.ts` lint debt at HEAD: `unwrapCauseError` cyclomatic complexity (11 > 8), `record-cast`, `manual-tagged-error`, 4 unused `_h/_t/_s/_sp`. These block the FINAL hook-passing commit and should be cleaned with the conformance work.

## Acceptance

- `nx run-many -t test test:conformance` green.
- The final commit passes the pre-commit hook (no `--no-verify`), which requires the `test-client.ts` lint debt cleared.

### LINKED ISSUE #740: fix(#705): app-client.test transport-resilience suite drives an app client with agent-callable methods
## Summary

On `impl/705-d-plus-c-cutover` (#722), `@moltzap/client:test` has **10 remaining failures**, all in `packages/client/src/app-client.test.ts`. Every other package's `test` + the full `test:conformance` + `test:integration` gates are green (verified on `4a54c4ec`).

## Root cause

These transport-resilience tests build a `MoltZapAppClient` (`makeClient`) and drive RPCs through `sendRpcEffect`, which forwards to `client.call(definition.name, params)`. `MoltZapAppClient.call<Tag extends AppCallableTag>` only dispatches methods in `AppCallableGroup` (`callablePrincipal: "app" | "any"`). The tests use AGENT-callable methods:

- `TaskList` (`task/list`, agent) — 7 sites
- `MessagesSend` (`messages/send`, agent) — 2 sites
- `AgentsLookupByName` (`agents/lookupByName`, agent) — 1 site

So `client.call("task/list", ...)` hits an empty slot in the app group's per-tag dispatch map → `map[tag] is not a function` (or the pending RPC never lands and the test times out / `WebSocket not connected`).

Failure modes observed: 4× `Test timed out`, 2× `map[tag] is not a function`, 2× `WebSocket not connected`, 1× `expected false to be true`, 1× `AsyncFiberException` on close.

## This is a test-design decision, not a mechanical fixture fix

Two valid resolutions, both touching `app-client.test.ts` + `app-client-test-support.ts`:

1. **Swap to an app-callable method** (e.g. `TaskClose`, `callablePrincipal: "app"`, params `{taskId}`, result `{task}`). Mechanical but broad: ~10 `sendRpcEffect` call sites, the mock-server response helpers (`sendMalformedFramesAndResponse`, `sendPaddedNotificationAndResponse`, `startAgentsLookupByNameServer` — they `encodeResponse` for `TaskList`/`AgentsLookupByName`), and the result/method assertions (`result.tasks` → `result.task`, `expect(captured.method).toBe(AgentsLookupByName.name)`). `TaskClose.encodeResponse` needs a valid `Task` fixture, unlike `{tasks: []}`.

2. **Use an agent client** for the transport tests. The file ALSO tests app-callback handling (`onAppCallback`, `dispatch/authorize`) which is app-specific, so this likely means splitting the transport-resilience tests into an `agent-client.test.ts` (none exists today) and keeping the app-callback tests here.

## What's already done on the branch

The mux-fixture half of `app-client-test-support.ts` IS fixed: the scripted server unwraps the `{ch,f}` envelope + strips native extras + stringifies numeric ids inbound, wraps outbound (responses c2s, notifications as void-result reverse RPCs on s2c). HelloOk + the app-callback scripted path work. Only the wrong-principal-method dispatch remains.

## Acceptance

`nx run @moltzap/client:test` green; `nx run-many -t test` green workspace-wide; the fix lands hook-clean (no `--no-verify`).

### LINKED ISSUE #741: transport: drop the {ch,f} mux envelope — route by @effect/rpc frame-family (post-#722)
## Finding

The `{ch,f}` mux envelope is **not strictly necessary** for routing. Verified (wf_bc47cc08-844) against @effect/rpc 0.75.1:

- The wire frame set partitions **totally and disjointly** into `FromClientEncoded` = {Request, Ack, Interrupt, Ping, Eof} and `FromServerEncoded` = {Chunk, Exit, Defect, Pong, ClientProtocolError}.
- Each physical endpoint stands exactly **one RpcClient + one RpcServer**, so inbound `FromClient`→ local RpcServer, inbound `FromServer`→ local RpcClient — deterministic, no ambiguous frame. Ping/Pong are direction-split; Eof is one-way; `ClientEnd` never serializes; there is no setup/handshake frame on the socket wire.

So frame-family alone routes every inbound frame. The `ch` tag's real value is **decoupling**: the mux treats the inner frame `f` as an opaque string and never parses it.

## Cost of the current design

That opacity forces **double-JSON-encoding**: `makeEnvelopeEncoder` does `JSON.stringify({ ch, f: wire })` where `wire` is already an encoded JSON string, so the inner frame is escaped-and-re-stringified, then double-parsed on receive.

## Proposed (post-#722, own pass)

Drop the envelope; route inbound frames by `_tag`/frame-family (`FromClient`→RpcServer, `FromServer`→RpcClient). Removes the double-encode + the wrapper.

**Required guard:** a compile-time exhaustiveness canary pinning the mux's `FromClient`/`FromServer` partition against @effect/rpc's `RpcMessage` unions, so a future @effect/rpc frame-set change fails the build instead of silently misrouting. This converts the only real downside (coupling the transport to @effect/rpc's wire vocabulary) into a compile-time-safe dependency.

## Scope / sequencing

- NOT for #722 (heading to stamina) or #728 (naming + group-collapse, not re-framing the wire). This is a wire-protocol change on a working system — its own pass with the canary + a full conformance/integration re-run.
- #728 still strips the vestigial `native-*` naming and **keeps `mux`** — confirmed a genuine two-engine multiplexer (c2s forward + s2c reverse for notifications/callbacks), so the name is earned.

### LINKED ISSUE #742: Descriptor modernization: `requires` unification + compile-soundness + by-method blocks
Post-#722 structural work (task #66). Full design in `.claude/simplification-audit/validated.json` + trial branch `explore/rpc-def-locality` @ `2cf4365`.

**Problem:** RPC definitions are hard to read/edit — organized by-KIND (errors/schemas/descriptors interleaved), and `callablePrincipal` + `caps` + `requiresActive` are three fields that are really one authorization spec. "caps" is also a misnomer (they're authorization checks that provide proof tokens, not object-capabilities).

**Locked decisions:**
- **Naming:** `caps` + `callablePrincipal` dissolve into ONE `requires: [...]` list (Effect-native — the handler's `R` channel *is* its requirements). Field `requires`, `RpcCapTag`→`Requirement`. Principal stays exactly-one-first (`AgentPrincipal`|`AppPrincipal`). `requiresActive` dissolves into a refinement requirement `AgentClaimed` (reads `connection.auth.agentStatus`, type-paired with `AgentPrincipal`).
- **Compile-soundness (both axes):** `requires`-required (TS2741 on omit, explicit `requires:[]` for `network/connect`, deletes the `?? ([] as unknown as)` cast) + handler-`E` ⊆ declared `errors` guard (#738; prereq: narrow the over-broad service-error aliases that erase to `~any`).
- **Readability:** by-method-block reorg WITHIN domain files (trial confirmed — NOT file-per-RPC); inline one-use schemas; the "Adding an RPC" guide rewritten (the old one referenced deleted symbols); split `app/methods.ts` (585L) → `manifest.ts`+`dispatch.ts`+`app-callbacks.ts`.

Trial caught a real doc-drift bug (phantom `RateLimitedError`). Wire impact: stable.

### LINKED ISSUE #746: Client cleanup from #722 review: typed E-channels, vestigial LogicalClock, lease audit, residual comments
From #722 review (chughtapan). Client-side cleanups the typed-error rearch enables. Lands on #722 (modernization, merge deferred).

1. **Use RPC typed E-channels; kill error-laundering casts.** The client launders RPC errors through casts (e.g. `channel-base/lease.ts:139` `return Effect.fail(err as E2)`) + regeneration. RPCs now carry tagged errors in the E channel — return the typed error directly, drop the casts. (`runtime/close-info.ts` same spirit; it's socket-level but clean it too.)

2. **LogicalClock is vestigial — remove it.** Audited: `observeMessage` (`channel-core.ts:638`) computes a logical clock (epoch + per-sender vector increment) and it's threaded through `clock` fields across client + server, but it is NEVER consumed — every `.clock` read is pass-through (`clock: x.clock`); no comparison/ordering/dedup uses the value. Computed-and-discarded. Remove `observeMessage` + the `clock` threading + the optional `clock` schema fields. Also audit the arena's lease/clock usage (flagged in review).

3. **Lease audit.** Review `channel-base/lease.ts` (client) + the arena lease code for other dead/erasure cruft.

4. **Residual comment cleanup** (the #733 capstone hit production files; these are client scripts/tests): `scripts/generate-cli-docs.ts`, `notification/__tests__/filter-equivalence.test.ts` (move to JSDoc / explain why), `notification/__tests__/snapshot-semantics.types-check.ts` (excessive comments + the `snapshot-semantics` name is unclear — clarify or rename).

### LINKED ISSUE #749: Cut the degenerate conformance model-equivalence oracle (K≈1, constant authorizationOutcome, unchecked ReferenceState)
## Smell
The conformance suite carries a "reference model" that is supposed to independently predict RPC outcomes and cross-check the live server — but it is **degenerate**, verifying almost nothing while looking like a both-sides oracle. Same category as the frame-mutator strict excess-key tier (cut in #741): test scaffolding that re-proves a triviality, not our behavior.

Evidence at the #722 tip:
- `testing/arbitraries/rpc.ts` → `confidentOracleMethods` is computed at module load and is **K≈1**; `arbitraryConfidentCall()` even guards `if (confidentOracleMethods.length === 0)` → throws "model has zero confident-oracle methods; flag needs-structural-rework". A property over ~1 method is not coverage.
- `testing/models/dispatch.ts` → `authorizationOutcome` is effectively constant (constant-deny for the unauth case it's used in); `applyCall` is a near-constant lookup, not a real reducer.
- `testing/models/state.ts` → `ReferenceState` fields are computed but never asserted against server state — an aspiration, not a checked contract.
- Consumers: `conformance/task/model-equivalence.ts` (the degenerate property), `conformance/identity/authority-negative.ts → assertModelDeniesUnauthenticated` (the live-server `assertTaskListDenied` in the same file is the REAL assertion and stays), `testing/__tests__/primitives.test.ts` (tests the model itself).

## Cut (verify-then-delete)
First VERIFY the degeneracy at HEAD (confirm K, confirm `authorizationOutcome` is constant for its call sites, confirm `ReferenceState` fields are never compared to server state). If the model genuinely checks real outcomes for K>1 methods, STOP — it's real coverage, do not delete.

If degenerate, delete: `testing/models/` (dispatch.ts, state.ts, index.ts), the confident-oracle exports in `arbitraries/rpc.ts` (`arbitraryConfidentCall`, `confidentOracleMethods`) + barrel, `conformance/task/model-equivalence.ts` + its registry entry, the model half of `authority-negative.ts` (`assertModelDeniesUnauthenticated` — KEEP `assertTaskListDenied`), and the model bits in `primitives.test.ts`. Keep all live-server conformance assertions.

## Sequencing
Touches the conformance tree (`authority-negative.ts`) that #741's driver rewrite also touches — run AFTER #741 lands to avoid collision.

chughtapan:
## Summary

Makes apps/TMs a first-class auth principal (`AuthenticatedContext = Agent | App`) and retires the in-process default-app loopback, then reorganizes the protocol + server packages around the documented domain surface and moves wire validation/dispatch onto `@effect/rpc`.

**Core cutover**
- **Apps as a first-class principal:** three-arm connection (`Unauthenticated | Agent | App`), the `appKey` connect path, HTTP `/api/v1/apps/register`, and the in-process loopback default-app retired in favor of a real registered app.
- **Package reorganization:** protocol restructured around domains (identity, network, task, conversation, message, socket, rpc, testing); server internals moved into matching domain folders — boot in `core/`, transport in `http/` + `socket/`, app-principal in `identity/apps/`.
- **Wire stack modernization:** validation migrated TypeBox + AJV → Effect Schema; client→server dispatch moved onto `@effect/rpc`; the `{ch,f}` mux envelope dropped for frame-family routing; a `requires`-based requirement model replaces caps / callablePrincipal / requiresActive.
- **Surface cleanup:** stale compatibility names + generated surfaces removed; protocol/server docs and generated module docs refreshed.

**Follow-on cleanups (this PR)**
- `/simplify` passes: shared `errorPayloadFields`, reverse-callback guard dedup, `connectPreamble`, concurrent conversation-list reads; plus a workspace-wide per-folder sweep (shared `serializeMoltZapProfileConfig`, table-driven `principalLayer`, shared `fanOut` / `guardCallback` / `sendFrame`, merged imports).
- **Build robustness:** `tsBuildInfoFile` co-located inside `dist`, so cleaning `dist` forces a correct rebuild — a stale root-level `tsconfig.tsbuildinfo` previously made `tsc -b` skip declaration emit and exit 0 with a broken `dist`.
- Dead `log_level` config knob removed; cold-reader READMEs added for `core/` / `http/` / `identity/apps/`; `ListRow → ConversationListRow`.
- **Import hygiene:** handlers stopped importing their own folder's barrel for service Tags (server file-level circular dependencies 18 → 14).
- `listGroupsEffect` now drains every `ConversationList` page (was silently truncating named groups past the first page).

## Closes

Closes #717
Closes #704
Closes #734
Closes #735
Closes #739
Closes #740
Closes #720
Closes #721
Closes #746
Closes #741
Closes #742
Closes #749
Closes #723
Closes #725
Closes #500
Closes #478
Closes #736
Closes #702

Also resolved or mooted by this branch and closed directly (with notes on each issue): #728, #751, #685, #214, #572, #582, #633, #615, #614, #611, #591, #681, #404, #454, #453, #211, #369, #333, #332.

## Kept open (follow-ups)

- **#724** — folder split delivered; what remains is ratifying the open-question resolutions the branch shipped (esp. app-principal Tag placement in `identity/apps/layer.ts` vs the recommended boot-side default).
- **#744** — barrel self-cycles fixed (18 → 14); the full server layer-DAG is a redesign of the barrel + composition-root pattern, not a cleanup, so it stays separate.
- **#737 / #733** — naming-convention long tail and the comment-lint capstone (a single 0-violation commit by design).
- **#402** — MoltZap-side lint gates + `@effect/rpc` descriptors done; cross-repo arena sub-tasks remain.
- Separate feature/perf/test work: **#743, #748, #750, #718, #639, #491, #670, #277, #474**.

## Verification

CI green on the latest head (`ci`, `conformance`, `test`, Mintlify deployment + link-rot). Every commit runs the pre-commit hook (`oxfmt`, `docs:generate`, docs drift/import/constant checks, `agent-code-guard-knip`, `pnpm check` = lint + format under `--max-warnings=0`, full `pnpm build`, type-safety + Effect-hygiene guardrails). Locally verified `nx run-many -t build` / `-t test` / `-t lint` green across all 7 projects.


chughtapan:
## STAGE 2 — Step 3 handoff: green-incremental decomposition (tree CLEAN at `f1982189`)

**Status:** `BLOCKED` (budget vs. atomic scope). No tree changes made; worktree `stage2-cutover` is clean at the green checkpoint `f198218957e057aea9cb3e6eca573db9fb5a5ade` (ci+test+conformance all `success`). Grep-state unchanged (decodeFrame/decodeFrames 21 files, closedStructGuard/STRICT_DECODE/decodesStrictly 12, originator 36 — all LIVE, as expected pre-cutover).

This session completed the full architectural read of the step-3 surface and confirms the path is sound (no architecture gap). The reason for handing off rather than pushing partial work: step 3 is a single atomic ~50-file cutover across 4 packages with TWO MED-risk type-reconciliation surfaces the spec itself flags as "where the cutover is most likely to not compile" (`ErasedSlot → HandlersFrom<R>` reshape; conformance-driver rewrite). Per-commit-green discipline (full pre-commit hook, no `--no-verify`) plus the prohibition on red/partial trees means this needs a session budget sized to the whole atomic core, not a partial.

### The irreducible atomic core (why no trivial green seam exists mid-reshape)

The handler BODIES live in 9 server `*.handlers.ts` authored via the `defineTaskMethod`/`defineAppMethod`/`defineNetworkMethod`(+`*Middleware`/`Connect`) family in `transport/define-layered-method.ts`, which today RETURN `ErasedSlot`. `RpcGroup#toLayer` requires `HandlersFrom<R>` = `{ [tag]: (payload, { clientId, headers }) => Effect<Success, Error, R> }` with `R` having the middleware `provides` excluded (verified against `@effect/rpc@0.75.1` `RpcGroup.d.ts → toLayer/HandlersFrom/HandlerContext`, `Rpc.d.ts → ToHandlerFn/ExcludeProvides`). Changing the definers' return type from `ErasedSlot` to a `HandlersFrom`-shaped body breaks all 9 handler files + `context.ts`'s `DispatchContext`/`defineMiddlewareMethod`/`defineUnauthMethod` + `app/server.ts buildServerSlots` simultaneously. That reshape cannot be made additive without forking the entire authoring layer.

### Verified API ground-truth (re-confirmed against installed `.d.ts` this session)
- `RpcGroup#toLayer<Handlers extends HandlersFrom<R>>(build): Layer<Rpc.ToHandler<R>, EX, Exclude<RX,Scope> | HandlersContext<R, Handlers>>` — per-tag totality is compiler-checked (`RIn ≠ never`), replacing `typed-dispatcher.types-check.ts`.
- `Rpc.ToHandlerFn = (payload: Payload<Current>, { clientId: number, headers: Headers }) => ResultFrom<Current,R>`. The handler is payload-only; principal/caps arrive via Context, NOT a `ctx` arg.
- `RpcMiddleware.TagClass` (class-form interface; `RpcMiddleware.Tag` is the constructor returning a `TagClass`). The middleware impl shape is `RpcMiddleware<Provides,E> = (options: { clientId, rpc, payload, headers }) => Effect<Provides, E>`.
- `CapabilityMiddleware` (`transport/capability-middleware.ts`) is moltzap's OWN cast-free per-cap metadata shape — KEEP it + the static `provideMiddleware` per-arm fold (§2e: do NOT route through `@effect/rpc`'s erased `applyMiddleware`). The `RpcMiddleware.TagClass` adoption in §2e is ONLY for the NEW principal/connection-resolution middleware (resolve `clientId → Connection` via `ConnectionManager`, narrow the 3-arm union, provide `CurrentPrincipal`/`ConnectionTag`).

### Proposed green-incremental commit plan (each a green, ff-pushed, CI-verified checkpoint)

**3a — additive native server-engine substrate (protocol).** Mirror step-2's additive pattern: a NEW protocol module that stands up `RpcServer.make(ServerRpcGroup)` + `RpcServer.layer` over the step-2 `native-mux` server protocol, plus the principal/connection-resolution `RpcMiddleware.TagClass` descriptor TYPE (resolves `clientId → Connection`, narrows to the 2-arm `Principal`, provides `CurrentPrincipal`/`ConnectionTag`). Add a types-check canary that a `HandlersFrom<ServerRpcGroup>` literal is accepted by `ServerRpcGroup.toLayer` and is per-tag-total. NO live wiring, NO `ErasedSlot` deletion. Green. (This de-risks the MED type surface in isolation before touching the live handlers.)

**3b — THE atomic reshape (protocol + server, one commit).** Reshape the definers to emit `HandlersFrom<ServerRpcGroup>` bodies instead of `ErasedSlot`: rewrite `context.ts` (`defineMiddlewareMethod`/`defineUnauthMethod`/`DispatchContext` → middleware-Context principal), `define-layered-method.ts` (all definers, KEEP the static `provideMiddleware` fold — only the OUTPUT shape changes), `middleware-slot.ts` output, all 9 `*.handlers.ts` (mechanical — bodies are unchanged, only the signature `(params, ctx)` → `(payload, {headers})` + principal via `CurrentPrincipal`), `app/server.ts buildServerSlots → ServerRpcGroup.toLayer(...)`. Switch `socket-handler.ts → handleFrame` onto the `RpcServer.Protocol` read loop over the native-mux socket (c→s channel), preserving connection-lifecycle/presence/lease/hook machinery; wire socket-close → `disconnects` Mailbox + per-client cleanup (native-mux's `runMuxReader` feeds this); readiness Deferreds not sleeps; re-add keepalive. `RpcClient.make(ServerRpcGroup)` for the client outbound (`agent-client.ts`, `service.ts:631` validateParams removed). Delete `typed-dispatcher.types-check.ts` (totality now native). Migrate the conformance `_shared/driver/test-client.ts`+`test-server.ts` onto the native Protocol substrate IN this commit (irreducibly atomic with the switch). Green. This is the heavy commit.

**3c — role-inverted callback (s→c).** Server holds `RpcClient<AppCallbackRpcGroup>`, client holds `RpcServer<AppCallbackRpcGroup>` over the native-mux s2c channel. Re-point `app-host.ts → callAppRpc` to `cbClient.client.<Method>(...)`; re-point `transport/connection.ts → sendRpcToClient`/`acquireConnectionRpcClient`/`type Originator`; `app-client.ts AppCallbackHandlers` → native server handler map. STOP USING `originator.ts` (do NOT delete — that's step 5). Complete the dual-endpoint demux. Re-point `RpcServerError` construction onto the native client failure channel (keep openclaw-channel's import working). Green.

If 3b genuinely exceeds a session even after 3a de-risks it, 3b is itself the natural escalation unit — it is the atomic reshape and cannot be split green.

### Steps 4-6 (unchanged from brief; after step 3 green)
4. Notifications → `@effect/rpc` streaming (~15 `defineNotification` → `Rpc.make({stream:true})`; `runtime/subscribers.ts` consumes a Stream).
5. DELETE dead wire + strict machinery + `originator` + frame schemas + `method.ts` dispatcher + MANDATORY docs-walker edit (`scripts/docs/schema.ts RPC_DEFINITION_FIELDS` + `types.ts AnyRpcDocDefinition`). Grep-zero after: decodeFrame/decodeFrames, closedStructGuard/STRICT_DECODE/decodesStrictly, originator → 0.
6. Residual consumers + perf re-baseline + openclaw-channel sweep. Honor #728 PIN-don't-MOVE + #731 (no tsconfig exclude mid-cutover).

**Recommendation:** resume with a session budget sized to 3a→3c (the brief's "full context budget"); 3a first to retire the MED type-risk in isolation, then the atomic 3b.

chughtapan:
## Integration suite green; unit `test` + `conformance` are a pre-existing wall

**`test:integration`: 142 passed / 0 failed** (was 13 failed). Commit `e7ad6ed7`.

Four real defects the native-mux + typed-error rearch surfaced, all fixed:

1. **Handler-raised-error-not-in-union (class bug).** 8 task-admin RPCs raise `ForbiddenError` via `assertCallerAppOwnsTask` but didn't declare it → engine couldn't encode → die → `InternalError`/null. Added `ForbiddenError` to all 8 unions. (presence/subscribe was the first instance, fixed earlier.) Compile-time guard for this class filed as #738 (the obvious canary is vacuous — engine member error is type-erased to `AnyNoContext`; the sound guard needs the over-broad service-error aliases narrowed first).
2. **Request-id overflow → concurrent-response misrouting.** Driver base `Date.now() * 1_000_000` ≈ 1.78e18 > `MAX_SAFE_INTEGER`; jsonRpc round-trips ids through `Number()`, so concurrent ids collapsed and responses misrouted → RPC timeouts. Dropped the multiplier. **This was the multi-agent messages/send hang.**
3. **Notification ack hang.** Server-originated notifications are void-result s2c RPCs awaiting the client ack; the driver demoted them to fire-and-forget → server's `notify` parked → dispatch deny-removal never ran. Driver now keeps the id, fans out to subscribers, AND acks.
4. **Mux dropped malformed frames** instead of replying `-32700`. `routeInbound` now replies parse-error (id:null) for undecodable chunks when a `reply` writer is wired (the live socket-handler wires it).

Plus stale-test fixes (`.tag` not `.code`; auth-failure prefixed credentials).

### Pre-existing / rearch-introduced

All four bugs are **rearch-introduced** (the native-mux + typed-error cutover; the driver never spoke the native wire before). Verified against `main` where applicable.

### The remaining wall (NOT introduced here)

`@moltzap/server-core:test` (**36 failed**) and `test:conformance` (**27 failed** server-side) are **pre-existing on this branch** — identical counts with and without `e7ad6ed7` (stash-and-compare verified). Root cause: the rearch routes notification fan-out through the s2c originator (void-result RPC awaiting ack), but the unit mocks (`presence.service.test`, `app-host.test`) record only `socket.write` and the conformance proof harness's property timeout budgets fire before detection. Tracked as **#739** (also covers the pre-existing `test-client.ts` lint debt that blocks the final hook-passing commit, and `native-mux.test.ts`'s 2 encoder failures).

chughtapan:
## Gate status update — conformance + most of the test gate green

Three hook-clean commits landed (no `--no-verify`): `e7ad6ed7`, `5a6a4e72`, `4a54c4ec`.

### Green now
- `@moltzap/protocol:test` — 219/0 (native-mux, test-client.app-callback, request-well-formedness flake killed)
- `@moltzap/server-core:test` — 186/0 · `test:integration` — 142/0
- **All 5 `test:conformance` suites** — 49 properties + 44 divergence proofs, 0 failed (dispatch-admission 1/16→16/16, server-executable 7/18→18/18, presence/delivery capture-scan fixed)
- `@moltzap/claude-code-channel` 45/0 · `@moltzap/runtimes` 40/0 · `@moltzap/openclaw-channel` 62/0 · `@moltzap/nanoclaw-channel` 20/0
- `@moltzap/client`: send CLI, service-socket-path, live-transport dedup — all green

### Root cause (all of the above)
The native-mux + typed-error rearch moved notification fan-out onto the s2c reverse RPC (`originator.notify`, void-result, awaits ack), wrapped the socket in a `{ch,f}` envelope, and migrated consumers to `service.call(tag, payload)`. The unit/conformance/channel harnesses still spoke the old bare-frame / `sendRpc` / direct-write wire, so they observed zero notifications or couldn't dispatch. Each was fixed at the right altitude (recording originator that encodes+writes; capture-scan that recognizes notification-as-s2c-request gated by the `notificationDefinitions` allowlist; mux-enveloped mock servers; `service.call` fixtures; `ForbiddenError(LeaseInvalid)` lease shape).

### Remaining (filed #740)
`@moltzap/client:test` has 10 `app-client.test` failures: the transport-resilience suite drives a `MoltZapAppClient` with AGENT-callable methods (`TaskList`, `MessagesSend`, `AgentsLookupByName`) the typed `call<AppCallableTag>` cannot dispatch. The mux-fixture half is fixed; what remains is a test-design decision (swap to an app-callable method like `TaskClose` + build the result fixture, OR split transport tests onto an agent client). Scoped in #740.

### Also filed earlier
- #738 — compile-time guard for handler-error ⊆ method-union (the obvious canary is vacuous; needs service-error-alias narrowing first)
- #739 — the unit/conformance redness root cause (now largely closed by the commits above)

PR #726, opened 2026-05-30
perf(wire): decode hot-path benchmark (de-risks #723 Effect-Schema migration)

chughtapan:
## What

Adds an inbound **wire-decode micro-benchmark** over real RPC frames that measures the full per-frame decode tax the dispatcher pays, and A/Bs it against an Effect-Schema prototype to answer the one open perf risk gating the #723 migration.

The decode tax is two stages, both timed against the **real shipping schemas** (no toy schemas):
1. `decodeFrame` (`wire.ts`) — the 3-arm envelope classifier (compiled AJV `validateRequestFrame` → `validateResponseFrame` → `validateNotificationFrame`).
2. `decodeRpcRequest` (`rpc-groups.ts`) — the linear `.find()` method-select over the live 31-method `serverRpcMethods` catalog + that method's compiled AJV `validateParams`.

Frames span three classes: **small** (`agents/list`), **typical** (`messages/send`, branded UUIDs + text-part array), **nested** (`task/conversation/create`, branded UUID + `participants[]`). Decode is isolated from socket I/O and JSON parsing (pre-parsed plain objects).

## Why (de-risks #723)

The [#720 spike](https://github.com/chughtapan/moltzap/issues/720) flagged per-frame `Schema.decode` (AST-interpreted) vs AJV (compiled) as the **one UNPROVEN risk** on this hot path. The [#723 P6 gate](https://github.com/chughtapan/moltzap/issues/723) is locked at **≤5× AJV / ≤50µs per frame**. This PR answers it with numbers.

The Effect-Schema prototype is built in the **exact shape the migration would ship**: a `Schema.Union` of per-method structs discriminated on the `method` literal, decoded via `Schema.decodeUnknownEither(..., { onExcessProperty: "error" })` — `onExcessProperty: "error"` matches the shipping `new Ajv({ strict: true })` + `additionalProperties: false` (effect's default silently strips excess keys; see #723 r4 B1). The union has exactly 31 arms = the live `serverRpcMethods` size, so the method-selection cost is a fair A/B. The 3 measured methods carry full param fidelity (branded-UUID `Schema.pattern(UUID_RE)`, nested arrays); the other 28 are fillers.

## Numbers (mean ns/decode, vitest bench, ~150k-260k samples/frame, rme <5%)

**Full inbound decode** — `decodeFrame` + `decodeRpcRequest` (the number that gates the hot path):

| Frame | AJV (shipping) | Effect Schema (proto) | Ratio | Schema absolute |
|---|---|---|---|---|
| small · agents/list | ~1.9 µs | ~3.6 µs | **1.9×** | 3.6 µs |
| typical · messages/send | ~2.6 µs | ~8.5 µs | **3.2×** | 8.5 µs |
| nested · task/conversation/create | ~2.4 µs | ~8.0 µs | **3.3×** | 8.0 µs |

Schema p99 tails stay ~20-30µs worst-case, still under the 50µs ceiling.

**Method-selection only** (AJV linear `.find()` + `validateParams` vs Schema union arm-select): AJV ~0.7µs, Schema ~8-9µs. The high ratio here is structural — AJV's `.find()` is a trivial string scan while Schema has no separate "select then validate" stage; the union decode does method-select AND full params validation in one walk. The composite (full inbound decode) is what matters and is the table above.

## Verdict: ACCEPTABLE

Schema's full per-frame decode is **~1.9-3.3× AJV, worst-case ~8.5µs/frame** — inside #723's locked P6 gate (≤5× AJV / ≤50µs) with large headroom. At 8.5µs/frame the decode stage alone tops out near ~118k frames/sec single-threaded before decode becomes the bottleneck — orders of magnitude above any realistic JSON-RPC frame rate moltzap sees. **Perf is not a blocker for the #723 migration**; no mitigation (precompiled Schema, ParseResult caching) is needed at current frame rates. The migration is de-risked on this axis.

## What this PR actually merges

The **AJV-baseline benchmark is the mergeable artifact**: a standing perf canary on the decode hot path, useful regardless of #723, and the baseline the migration must not regress.

- `wire-decode.bench.ts` — the harness (AJV + Schema variants).
- `frames.ts` — real fixtures + the exact shipping AJV decode composition.
- `frames.test.ts` — runs in the normal suite; asserts both engines accept all three fixtures AND that the prototype union arm count tracks `serverRpcMethods`, so the canary can't silently drift.
- `effect-schema-prototype.ts` — clearly-fenced throwaway spike code (not on any barrel, not imported by production), kept as a bench variant so the A/B is reproducible.

Bench files live under `src/**/__benchmarks__/`, **excluded from the published build** (`tsconfig.json`) so nothing ships to `dist/`, but **typechecked** via `tsconfig.bench.json` (`pnpm bench:typecheck`) and run via `pnpm bench`. Registered as a knip entry.

## Gates

`pnpm check` (lint + format), `pnpm build`, `pnpm typecheck`, `pnpm bench:typecheck`, sloppy-code-guard, knip, oxlint, and the full protocol suite (220 tests) all green. Pre-commit hook ran clean (no `--no-verify`). Ran `/simplify` + `/review` on the diff; review caught and fixed a union-arm-count drift (prototype had 35 arms vs the real 31) and added the parity test that locks it.

Refs #723, #705, #720

### DIFF
--- a/knip.json
+++ b/knip.json
@@ -51,6 +51,7 @@
       "entry": [
         "src/**/*.test.ts",
         "src/**/*.types-check.ts",
+        "src/**/__benchmarks__/*.bench.ts",
         "src/testing/conformance/**/*.ts",
         "scripts/check-mermaid.ts",
         "vitest*.config.ts"
--- a/packages/protocol/package.json
+++ b/packages/protocol/package.json
@@ -48,7 +48,9 @@
     "lint": "eslint src scripts *.ts --cache --max-warnings=0",
     "docs:generate": "tsx scripts/generate-docs.ts && (cd ../.. && pnpm exec typedoc) && tsx scripts/generate-modules.ts",
     "test": "vitest run",
-    "test:watch": "vitest"
+    "test:watch": "vitest",
+    "bench": "vitest bench --run src/transport/__benchmarks__/wire-decode.bench.ts",
+    "bench:typecheck": "tsc --noEmit -p tsconfig.bench.json"
   },
   "dependencies": {
     "@effect/platform": "^0.96.1",
--- a/packages/protocol/src/transport/__benchmarks__/effect-schema-prototype.ts
+++ b/packages/protocol/src/transport/__benchmarks__/effect-schema-prototype.ts
@@ -1,0 +1,201 @@
+/**
+ * @file Effect-Schema PROTOTYPE for the #723 wire-validation migration.
+ *
+ * THROWAWAY SPIKE CODE — bench-only. Not on any barrel, not imported by
+ * production. It exists to A/B the per-frame decode cost of `effect/Schema`
+ * (AST-interpreted) against the shipping TypeBox+AJV (compiled) path, in the
+ * EXACT shape #723 would ship per the #720 spike verdict:
+ *
+ *   - one `Schema.Struct` per wire method, discriminated on the `method`
+ *     literal (the migration's `Match.discriminator("method")` ground), with
+ *     the same `jsonrpc`/`id`/`method`/`params` envelope the real
+ *     `RequestFrameSchema` carries;
+ *   - all ~34 server-inbound methods folded into one `Schema.Union` so the
+ *     union arm-select cost matches the real catalog the AJV `.find()` scans;
+ *   - decoded via `Schema.decodeUnknownEither(Union, { onExcessProperty:
+ *     "error" })` — `onExcessProperty: "error"` is REQUIRED to match the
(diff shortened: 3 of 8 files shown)


chughtapan:
Closing without merging — the benchmark served its purpose as the #723 Half-2 perf spike (verdict ACCEPTABLE: Schema decode ~1.9-3.3x AJV, ~8.5µs worst, well inside the <=5x/<=50µs gate; full numbers on #723 comment 4581078189). Per decision, we're not checking in a standing benchmark. The #723 migration impl will do a one-off perf re-measure to confirm it stays under the gate, rather than maintaining a committed canary.

PR #949, opened 2026-08-04
docs(harness): reconcile implementation slate with lease-free production

chughtapan:
## What changed

- Reconciles the non-normative Harness implementation slate with lease-free production after #941 and the daemon receive/reply foundation in #942.
- Separates landed production from open candidates #943 through #948 and from the target state.
- Records participant enrichment as loopback-MCP-only presentation data passed to the HarnessClient implementation; the canonical production Conversation and main WebSocket wire remain unchanged.
- Records the still-unselected production profile-to-MCP acquisition and durable checkpoint composition instead of inventing them.
- Makes the clean-slate prerequisites explicit: v2/transcript and v2/harness remain scaffolds, and Transcript/Ledger plus the remaining management, ingress, HarnessClient, checkpoint, and plural-action contracts must be admitted before their corresponding implementation slices.
- Flags existing v2 authority drift around production leases and registration recovery without changing an ADR or normative spec.

## Why

The previous slate still described dispatch leases, conversation_busy, server grants, duplicate-reply suppression, and selected production registration recovery. Those claims no longer match current main or do not have a main-owned contract. The implementation handoff needs to describe current evidence without silently turning unresolved mechanics into requirements.

## Scope

This is an implementation-state and sequencing update only. It does not modify an ADR, normative specification, main-wire schema, runtime code, or clean-slate contract.

## Validation

- `pnpm nx run workspace:precommit`
- `pnpm docs:check:drift`
- `git diff --check`
- Independent factual review, followed by a clean re-review after both findings were corrected

## Related

- Tracks #926
- Production candidate stack: #943, #944, #945, #946, #947, #948
- Upstream v1 control-plane simplification: https://github.com/chughtapan/moltzap/pull/924

### DIFF
--- a/docs/architecture/components.md
+++ b/docs/architecture/components.md
@@ -1,6 +1,6 @@
 # Components
 
-Status: GATE 1 CANDIDATE — BLIND REVIEW REQUIRED
+Status: GATE 1 AUTHORITY REVIEWED — IMPLEMENTATION READINESS VARIES BY OWNER
 
 Decision owners:
 [`20260728-six-deep-packages-one-version.md`](../decisions/20260728-six-deep-packages-one-version.md),
--- a/docs/architecture/first-implementation.md
+++ b/docs/architecture/first-implementation.md
@@ -2,7 +2,7 @@
 
 {/* @bake-constants: V2_PROTOCOL_VERSION */}
 
-Status: **HARNESS AUTHORITY CANDIDATE — BLIND REVIEW REQUIRED BEFORE HARNESS CODE**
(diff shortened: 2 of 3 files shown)


chughtapan:
Superseded by the four-layer cutover. Its principal target, `docs/architecture/harness-implementation-slate.md`, was deleted along with the seven other cutover-archaeology files; the surviving orientation is `components.md`, `first-implementation.md` and `layers.md`. There is no file left for this reconciliation to apply to. Reopen against the surviving docs if the lease-free authority wording is still missing there.

PR #1026, opened 2026-09-03
fix(simulator): leave OpenClaw's a2a channel out of the message tool

chughtapan:
With the bundled Agent2Agent plugin enabled beside the MoltZap channel adapter, OpenClaw's `message` tool offers two channels for the same peers. A model that reaches for `a2a` first fails (`Unknown target "agent:sarah-agent" for A2A`), retries on `a2a` with the bare name, and either finds MoltZap on the third attempt or never sends at all.

Seen in the propagation bench with one agent on `anthropic/claude-opus-4-8` under 2026.902.1: one run went the whole observation window without a message, the next recovered on the third attempt. gpt-5.4 agents pick MoltZap every time, which is why it never showed before.

Every simulator peer is reachable through MoltZap alone, so the OpenClaw config now disables the `a2a` plugin entry. Test updated; `configuration.test.ts` and `runtime.test.ts` pass.

https://claude.ai/code/session_019tUudcav5hZzcPRLxf6J6X

### DIFF
--- a/packages/simulator/src/agents/openclaw/configuration.test.ts
+++ b/packages/simulator/src/agents/openclaw/configuration.test.ts
@@ -88,12 +88,15 @@ describe("buildOpenClawConfig", () => {
     assert.isUndefined(mcpSection(undefined));
   });
 
-  it("loads and enables the mounted channel adapter", () => {
+  it("loads the mounted channel adapter and disables the built-in a2a channel", () => {
     assert.deepStrictEqual(openClawConfig(undefined).plugins, {
       load: {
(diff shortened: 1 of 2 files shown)


chughtapan:
Closed as already landed. The same change is on main as 01687c35 (2026-09-03, same title): `pluginConfiguration` in `packages/simulator/src/agents/openclaw/configuration.ts` sets `[OPENCLAW_A2A_PLUGIN_ID]: { enabled: false }`. This branch conflicts only because main already carries it.

